PR #15, opened 2017-11-03
Initial Unit Tests for platform_login/platform_logout (and some templates)

DaemonWill:
### Description of the Change

Generated initial unit tests for the platform_login/logout repositories given the following scenarios: 

**Login**
- username is not provided
- password is not provided
- http method other than POST was used with handler
- valid information was passed and sent to cognito

**Logout**
- access token/authorization isn't provided
- valid information was passed to get the user's data using CognitoIdentityService
- given the above^ globalsignout is called with CognitoIdentityService

Provided updated dependencies and necessary additional files for configuration/running tests. (should do an npm install before testing if your environment is not updated)

Trailing unit tests for some java-templates will be included from previous work.

### Benefits

Ability to check if expected functionality works given new code changes

### Possible Drawbacks

expected returned values, results, and/or parameters may be added later which would mean tests should be updated.


api-template-java/src/test/java/com/slf/services/HandlerTest.java
```diff
@@ -17,23 +17,116 @@
 package com.slf.services;
 
 import java.util.HashMap;
+import java.util.Map;
```
suryajak: This is our template file that gets committed to the user repo when they create a Java service, so it wouldn't make sense to add unit tests that would fail once the user starts changing code.

lambda-template-java/src/test/java/com/tmobile/services/HandlerTest.java
```diff
@@ -14,23 +14,41 @@
 // limitations under the License.
```
suryajak: This is our template file that gets committed to the user repo when they create a Java service, so it wouldn't make sense to add unit tests that would fail once the user starts changing code.

platform_login/components/error-handler.js
```diff
@@ -22,40 +22,40 @@
 **/
 
 module.exports = (logger) => {
-    
+
 	var errorObj = {
-        throwInputValidationError: function(errorCodeNum, errorMessage) { //Raise a  bad requests exception
-            return {
-                errorCode: errorCodeNum.toString(),
```
suryajak: Why are we reverting this code?

platform_login/test/test.js
```diff
@@ -15,13 +15,102 @@
...
+  * Given a valid event object, handler() should invoke cognito's authenticateUser()
+  */
+  it('should pass in user data for authentication', function() {
+    //mocking "authenticateUser" function
+    stub = sinon.stub(AWSCognito.CognitoUser.prototype, "authenticateUser", spy);
+    var returned = index.handler(event,context);
+    stub.restore();
+    assert.isTrue(spy.called);
+    //expect(() => index.handler(event,context)).to.throw();
   });
```
suryajak: One more test case to handle invalid username/password combo

platform_login/test/test.js
```diff
@@ -15,13 +15,102 @@
...
+  /*
+  * Given an event with no username, handler() shows that a 101 error has been made
+  * @param {object} event without username information
+  * @param {object} aws context
+  * @param {function} callback function that returns what was passed
+  * @returns {string} callback function showing username was not provided
+  */
+  it("should throw a 101 error for missing username", function(){
+    event.body.username = undefined;
+    var bool = index.handler(event,context,callback).includes("101");
```
suryajak: We also need to assert that we have a non-empty error message for all the failure scenarios as this is used by UI to display errors.

platform_logout/test/test.js
```diff
@@ -15,13 +15,70 @@
...
+    AWS.mock("CognitoIdentityServiceProvider","getUser", (paramss, cb) => {
+      cb(null, "successfully identified user");
+    });
+    AWS.mock("CognitoIdentityServiceProvider","globalSignOut",spy);
+    var returned = index.handler(event,context,callback);
+    AWS.restore("CognitoIdentityServiceProvider");
+    assert.isTrue(spy.called);
+    //expect(() => index.handler(event,context)).to.throw();
   });
 });
```
suryajak: Add a test case where cognito fails to look up user

platform_login/test/test.js
DaemonWill: I will include this scenario; however, with the code in the index.js, once the username is found invalid an error will be thrown indicating to the user they're missing a username. Logic to validate the password wouldn't be called once the username is found invalid, so for this scenario I will be asserting that a 101 error is thrown and that a 102 was not (which may seem a bit redundant).

suryajak: user name is valid, password is valid (that is they are provided), but the combo might be wrong. [This line](https://github.com/tmobile/jazz/blob/master/platform_login/index.js#L83) is what the test case would validate.

DaemonWill: Ok, will include that case as well.

suryajak:
Thanks for adding these unit tests

PR #29, opened 2017-11-21
CAPI-907 Unit tests for platform-service GET functionality

DaemonWill:
### Requirements

Problem Statement: Add unit tests for platform_services for GET.
Acceptance Criteria:

1. the GET part of the REST API of platform_services will have 90%+ code coverage

### Description of the Change

Integrated tests for get-method processing in platform-service, added in dependencies and object data needed for test setup.

### Benefits

Cleaner future code development. Changes to existing logic can be checked in order to ensure the rest of the code base is unaffected.  

### Possible Drawbacks

dramatic changes to functional outputs and parameters can cause tests and potential build processes to fail.


platform_services/test/test.js
```diff
@@ -13,13 +13,200 @@
...
+    var logResponse = logStub.args[0][0];
+    var cbResponse = stub.args[0][0];
+    var logCheck = logResponse.includes(logMessage) && logResponse.includes(err.errorType) &&
+                    logResponse.includes(err.message);
+    var cbCheck = cbResponse.includes(errType) && cbResponse.includes(errMessage);
+    AWS.restore("DynamoDB");
+    logStub.restore();
+    stub.restore();
+    assert.isTrue(logCheck && cbCheck);
+  });
```
suryajak: This is a great start. Can you add a unit test case where "admin_users" can query all services? Similarly the second case where the dyanamodb table has records from multiple users and the GET only gets the specified users records.

suryajak: We need unit tests for this query param "last_updated_after" and "last_updated_before"

DaemonWill: Ok I see it in the getList.js, I'll include that as well.

platform_services/test/test.js
```diff
@@ -209,4 +342,45 @@ describe('platform_services', function() {
     stub.restore();
     assert.isTrue(logCheck && cbCheck);
   });
```
suryajak: The changes look good for GET. Can you update this PR to submit to v1.1.1 (v1.1 is already merged into master and released)?

PR #44, opened 2017-12-06
CAPI-962 Unit tests for platform_logs

DaemonWill:
### Requirements

tests for platform_logs, ~90% coverage

### Description of the Change

~97% coverage with unit tests for platform_logs

### Benefits

check future code changes against existing functionality

### Possible Drawbacks

changing functional parameters or outputs can cause tests to fail which may cause optional builds to fail

### DIFF
--- a/platform_logs/test/test.js
+++ b/platform_logs/test/test.js
@@ -1,11 +1,239 @@
-const assert = require('chai').assert;
+const chai = require('chai');
+const assert = chai.assert;
+const expect = chai.expect;
 const index = require('../index');
+const awsContext = require('aws-lambda-mock-context');
+const sinon = require('sinon');
+const logger = require("../components/logger.js");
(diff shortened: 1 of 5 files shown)


platform_logs/test/test.js
```diff
@@ -1,11 +1,357 @@
...
+      "body_put":{
+        "endpoints" : {
+          "dev": "newval"
+        }
+      },
+      "body":{
+        "environment": "test",
+        "service": "dAnCe",
+        "domain": "bA11r00m",
+        "category": "api",
```
suryajak: category is also a variable to this API. One of the other values is function, so please add test to validate that.

suryajak: Also looks like we have a bug where our code moves forward when it is none of those two values, so please add a unit test for that too.

PR #57, opened 2017-12-14
CAPI1010 addressing Issue50

DaemonWill:
### Requirements

Upon login and entering the services page, admin is able to see all of the services in the database regardless of which user created the service. For this scenario, only grant the admin access to the data for the services created by them, and not by other users. 

### Description of the Change

Current fix applies a db search filter for services created by the current user if the original http method of the request was a "GET," this includes admin users. Admin maintenance of the app/services for updating, creating, and deleting should still be in place unless they are also called by "GET."

### Benefits

TODO

### Concerns

Could affect Jenkins builds concerning copying, deleting, or updating existing services.

### DIFF
--- a/platform_services/components/crud/getList.js
+++ b/platform_services/components/crud/getList.js
@@ -23,6 +23,7 @@
 **/
 
 const utils = require("../utils.js")();
+const _ = require("lodash");
 
 module.exports = (query, getAllRecords, onComplete) => {
     // initialize dynamodb
@@ -83,7 +84,7 @@ module.exports = (query, getAllRecords, onComplete) => {
 		});
     }
 
-    if (!getAllRecords) {
+    if ((global.userId && !_.includes(global.config.admin_users, global.userId.toLowerCase())) || !getAllRecords) {
         var ddb_created_by = utils.getDatabaseKeyName("created_by");
 
         // filter for services created by current user
--- a/platform_services/components/crud/getList.js
+++ b/platform_services/components/crud/getList.js
@@ -25,7 +25,7 @@
 const utils = require("../utils.js")();
 const _ = require("lodash");
 
-module.exports = (query, onComplete) => {
+module.exports = (query, method, onComplete) => {
     // initialize dynamodb
     var dynamodb = utils.initDynamodb();
 
@@ -84,7 +84,7 @@ module.exports = (query, onComplete) => {
 		});
     }
 
-    //if (global.userId && !_.includes(global.config.admin_users, global.userId.toLowerCase())) {
+    if (global.userId && !_.includes(global.config.admin_users, global.userId.toLowerCase()) || method === 'GET') {
         var ddb_created_by = utils.getDatabaseKeyName("created_by");
 
         // filter for services created by current user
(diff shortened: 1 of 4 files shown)


platform_services/components/crud/getList.js
```diff
@@ -22,10 +22,10 @@
 	@version: 1.0
 **/
 
-const utils = require("../utils.js")(); 
+const utils = require("../utils.js")();
 const _ = require("lodash");
 
-module.exports = (query, onComplete) => {
+module.exports = (query, method, onComplete) => {
```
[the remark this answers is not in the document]
DaemonWill: That's doable; however, I'll be assigning that boolean not only on the user's status as an admin, but also based off of the method used. We currently have logic to check for the user's admin status and filtering records accordingly so I'm thinking this additional criteria for the boolean is needed, my only concern is if there are any jenkins builds that will require an admin to do crud.getList to retrieve all the services before an update/delete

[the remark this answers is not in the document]
DaemonWill: I'll make the change first and update the comment accordingly to addresses the potential concern.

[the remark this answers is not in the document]
DaemonWill: parameter and logic is in place, the additional criteria being checked for is the event.method on top of admin status. Does this seem like a valid approach?

platform_services/components/crud/getList.js
```diff
@@ -80,11 +79,11 @@ module.exports = (query, onComplete) => {
 				attributeValues[(":" + key_name)] = {
 					'S': query[key]
 				};
-			}   
+			}
 		});
     }
 
-    if (global.userId && !_.includes(global.config.admin_users, global.userId.toLowerCase())) {
```
suryajak: This logic should reside in the crud only. The consumer of the crud can ask for all records, but if they are not the admin(s), it should just return the current logged in user records.

platform_services/index.js
```diff
@@ -320,7 +332,8 @@ module.exports.handler = (event, context, cb) => {
                 },
                 // Check if a service with same domain and service_name combination exists
                 validateServiceExists: function(onComplete) {
-                    crud.getList({ 'service': service_data.service, 'domain': service_data.domain }, function onServiceGet(error, data) {
+                    getAllRecords = getAllCheck(event.method);
+                    crud.getList({ 'service': service_data.service, 'domain': service_data.domain }, getAllRecords, function onServiceGet(error, data) {
```
suryajak: Here it should say: crud.getList({ 'service': service_data.service, 'domain': service_data.domain }, getAllRecords: true,

platform_services/components/crud/getList.js
DaemonWill: Ok that makes sense. Thanks for the clarification

platform_services/components/crud/getList.js
```diff
@@ -80,11 +80,11 @@ module.exports = (query, onComplete) => {
...
+    if ((global.userId && !_.includes(global.config.admin_users, global.userId.toLowerCase())) || !getAllRecords) {
```
suryajak: small optimization based on how OR works [short circuit eval](https://en.wikipedia.org/wiki/Short-circuit_evaluation) - you want to move !getAllRecords to the front

PR #60, opened 2017-12-18
CAPI1010 re-addressing issue50

DaemonWill:
### Requirements

Upon login and entering the services page, admin is able to see all of the services in the database regardless of which user created the service. For this scenario, only grant the admin access to the data for the services created by them, and not by other users. 

### Description of the Change

Current fix applies a db search filter for services created by the current user if the user is not listed as an admin or a query param wasn't passed to check for allServices. 

Added query params for requesting platform/services in relevant jenkins files, updated platform/services swagger to accept additional param, added contional logic in index.js for platform/services. The param being checked for is "allServices"

### Benefits

TODO

### Concerns

Only services created by admin users achieve an "active" state, services created by non-admins remain on "creation_completed"

### DIFF
--- a/delete-serverless-service-build-pack/Jenkinsfile
+++ b/delete-serverless-service-build-pack/Jenkinsfile
@@ -1043,7 +1043,7 @@ def loadEventsModule() {
 
 def loadServiceInfo() {
 	def token = setCredentials()
-	def url = "$g_base_url/platform/services?domain=$domain&service=$service"
+	def url = "$g_base_url/platform/services?domain=$domain&service=$service&allServices=true"
 	def curlOutput = sh (script: "curl -H \"Content-Type: application/json\" \
 			-H \"Authorization: $token\" \"$url\" | $jq -r '.data[0] | [.created_by] | join(\",\")'", returnStdout: true)
 	def values = curlOutput.split(',')
--- a/delete-serverless-service-build-pack/Jenkinsfile
+++ b/delete-serverless-service-build-pack/Jenkinsfile
@@ -1031,7 +1031,7 @@ def loadServerlessConfig(String runtime, String configPackBBURL) {
 
 def updateServiceInDB(service_id_in_db, devApiID, region, statusval) {
   sh "curl -H \"Content-Type: application/json\"  -H \"Authorization: "+params.auth_token+"\" -X PUT -k -v -d \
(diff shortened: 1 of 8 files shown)


delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -1043,7 +1043,7 @@ def loadEventsModule() {
```
suryajak: We need to update this in updateServiceInDB call too.

delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -1043,7 +1043,7 @@ def loadEventsModule() {
 
 def loadServiceInfo() {
 	def token = setCredentials()
-	def url = "$g_base_url/platform/services?domain=$domain&service=$service"
+	def url = "$g_base_url/platform/services?domain=$domain&service=$service&allServices=true"
```
suryajak: name this parameter as isAdmin

jenkins-build-pack-api/Jenkinsfile
```diff
@@ -789,7 +789,7 @@ def loadServiceInfo(service, domain, serviceUrl, endpointKey) {
 
 	setCredentials(g_base_url + '/platform/login')
 
-	def url = "$serviceUrl?domain=$domain&service=$service"
+	def url = "$serviceUrl?domain=$domain&service=$service&allServices=true"
```
suryajak: Same as the above two comments:
1. We need to update this in updateServiceInDB call too.
1. name this parameter as isAdmin

platform_services/index.js
```diff
@@ -60,6 +60,12 @@ module.exports.handler = (event, context, cb) => {
         global.services_table = config.services_table;
         global.userId = event.principalId;
         var getAllRecords;
+        if(event.query && event.query.allServices){
```
suryajak: name this parameter as isAdmin

platform_services/swagger/swagger.json
```diff
@@ -110,6 +110,11 @@
                     "in": "query",
                     "required": false,
                     "type": "string"
+                }, {
+                    "name": "allServices",
```
suryajak: name this parameter as isAdmin

PR #66, opened 2017-12-29
Platform email integration

DaemonWill:
### Requirements

Problem Statement: Jenkins file currently uses plugin for email. We want to move to use platform_email to send email notifications

platform_email requires authorization header, so login using admin credentials and then send the following in the post body:

{ "from": '"Jazz Admin" <adminemailaddress>', "to": "toaddress", "subject": "subject", "text": "text to be sent" }
Acceptance Criteria:

1. Notifications continue to work while Jenkins uses platform_email and not the plugins

### Description of the Change

Changes made to the jenkins build for api/lambda/website build-packs as well as service-onboarding/delete-serverless-service build-packs to allow for integration of the platform/email service.

### DIFF
--- a/platform_email/index.js
+++ b/platform_email/index.js
@@ -103,7 +103,7 @@ function validateInput(userInput) {
  */
 function sendEmail(config, userInput) {
 	return new Promise((resolve, reject) => { 
-		let transporter = nodemailer.createTransport({
+		var transporter = nodemailer.createTransport({
 			SES: new AWS.SES({
 				apiVersion: '2010-12-01',
 				region: config.REGION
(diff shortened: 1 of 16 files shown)


DaemonWill:
Changes made according to feedback

platform_email/index.js
```diff
@@ -16,7 +16,7 @@
 
 /**
 Jazz Email service
-@author: 
+@author:
```
suryajak: Why did this file become part of this PR?

DaemonWill: The file was most likely used for debugging and some white space may have been added by the code editor, these changes are negligible and can be reverted

PR #71, opened 2018-01-04
V1.1.3 event bug: Building the stack with same name the second time.

r-pai:
### Requirements

This is a bug, when building the stack with same name the second time after we delete the stack causes build issues. The platform services fails with the below error.

An error occurred (ResourceConflictException) when calling the CreateEventSourceMapping operation: The event source arn (arn:aws:kinesis:us-east-1:108206174331:stream/jazz20171210-events-hub-dev) and function (jazz20171210-hndlr-dev) provided mapping already exists. Please update or delete the existing mapping with UUID 106257e3-ec6a-4fb2-af2d-f26f163a603d

### Description of the Change

This is a bug building the stack with same name the second time after we delete the stack causes build issues. The platform services fails with the below error.

An error occurred (ResourceConflictException) when calling the CreateEventSourceMapping operation: The event source arn (arn:aws:kinesis:us-east-1:108206174331:stream/jazz20171210-events-hub-dev) and function (jazz20171210-hndlr-dev) provided mapping already exists. Please update or delete the existing mapping with UUID 106257e3-ec6a-4fb2-af2d-f26f163a603d

### Benefits

platform services would be build correctly when the stack is created with same name on same day

### Possible Drawbacks

NA

### Applicable Issues

NA


suryajak:
@r-pai : Can you take a look at the comment?

PR #73, opened 2018-01-06
Capi1039 Use External Module for Env.Vars

DaemonWill:
### Requirements
 The installer config has been transferred to a pre-configured jazz-installer-vars.json in the jazz-build-module repo in the SLF project. This task is to build an env-loader.groovy module that can provide the values from the jazz-installer-vars.json as the env.variables in future jenkins builds

Acceptance Criteria:

1. Jazz services (create/update/delete) continue to work post migration

### Description of the Change

Created an env-loader module in the jazz-build-module repository. Given a String that can be deconstructed into a json, the module will populate its own fields and have getters and setters acceptable for them. These fields will act as the env.vars used in various jenkinsFiles for several of the jazz directories. This PR shows logical integration for several of these files. 

### Benefits

Encapsulated data for sensitive values like the env.variables. 

### Possible Drawbacks

build failures due to all project folders having a dependency for this module. 

### Applicable Issues

Full integration away from calling upon the jenkins env.vars is not an easy task. In order to get the json, we need a pointer to the repo and credentials to fetch information, so we'll need at least the env.REPO_CREDENTIAL_ID, REPO_CORE, and REPO_BASE left out.. OR we can provide those values as temporary values passed as the build params. Please reach out if this comment is confusing.


jazz-build-module/env-loader.groovy
```diff
@@ -0,0 +1,158 @@
+#!groovy?
```
suryajak: Rename this file to config-loader

jazz-build-module/env-loader.groovy
```diff
@@ -0,0 +1,158 @@
...
+
+echo "the module, 'env-loader', loaded successfully... congratulations..."
+
+def initialize(installerVars){
+  if(installerVars != null){
+		def resultJson = parseJson(installerVars)
+
+    echo "assigning env values from installer..."
+
+    setCognito(resultJson.COGNITO)
```
suryajak: While I appreciate all the methods to map keys to methods, I believe this will cause a maintenance overhead with no additional value. 

Let this module return the complete JSON object which will be referenced in build packs as configLoader.AWS_CREDENTIAL_ID,  configLoader.API.API_ID_PROD and so on ...

DaemonWill:
changes made as requested

PR #87, opened 2018-01-16
Changing env dependencies for the configModule to params

DaemonWill:
### Requirements

Remove excess env.vars from all the jenkins builds and have them be implemented as params values where applicable.

### Description of the Change

Added in configLoader integration logic for all jenkinsFiles and passed in future params values as input values. 

### Possible Drawbacks

Integration with the installer has to be made

### Applicable Issues
The jazz installer project has yet to set these values, builds will fail unless updated on the installer side.


suryajak:
We are no longer taking this approach, so closing this PR.

PR #92, opened 2018-01-18
Capi1116 remove hard coded region from Jazz-core

DaemonWill:
### Requirements
The current installer has been validated to work in the us-east region only.

These are the places we know of so far:

cognito-authorizer\index.js

create-serverless-service\Jenkins

delete-serverless-service\Jenkinsfile

delete-serverless-service-build-pack\Jenkinsfile_Cleanup_CF_Dist

platform_email\Jenkinsfile_Platform

platform_logs\Jenkinsfile_Platform

platform_usermanagement\Jenkinsfile_Platform

Acceptance Criteria:

1. Parameterize the build files to take region from a pre-defined place (currently its env but will likely move to an object if  CAPI-1039 QA  is complete). - Pull Request 73
2. Parameterize all services code to use region from config files.
### Description of the Change

Added in the needed dependencies in the above mentioned files for including the region from the configuration json made upon installation. Re-assigned region calls to have that var value.

### Benefits

Region isn't hard coded and can be set upon installation/changed as seen fit.

### Possible Drawbacks

There may be a possibility that us-east-1 is set as a hardcoded value in more obscure parts of the code base.


DaemonWill:
PR updated with removed debug statements

PR #113, opened 2018-01-26
v1.3.1

suryajak:
### Description of the Change

Refactoring code to drive down the number of variables.

### Benefits

Smaller code base, improved readability and helps with merges.

### Possible Drawbacks

Functionality might get broken because of varied cases of variable declaration, initialization, and usage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ node_modules
 */target/
 .nyc_output/
 npm-debug.log
+*/package-lock.json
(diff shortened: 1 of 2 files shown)


suryajak:
I am not sure if it would be a configLoader thing. Do we expect this group/folder to change? Perhaps make it a variable as opposed to a hard coded string. We have an SCM module in PR, maybe it can be a variable in there.

PR #114, opened 2018-01-26
:fire: cleanup up unused variables and use configLoader reference for…

suryajak:
### Description of the Change

Refactoring code to drive down the number of variables.

### Benefits

Smaller code base, improved readability and helps with merges.

### Possible Drawbacks

Functionality might get broken because of varied cases of variable declaration, initialization, and usage.

### DIFF
--- a/jenkins-build-pack-lambda/Jenkinsfile
+++ b/jenkins-build-pack-lambda/Jenkinsfile
@@ -50,15 +50,6 @@ def Event_Status = [
 
 @Field def configModule
 @Field def configLoader
-@Field def var_api_key
-@Field def var_api_id_dev
-@Field def var_aws_credential_id
-@Field def var_env_name_prefix
-@Field def var_repo_core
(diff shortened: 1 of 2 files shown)


jenkins-build-pack-lambda/Jenkinsfile
```diff
@@ -67,44 +58,29 @@ def Event_Status = [
...
+	def _event = ""
 
     def serviceName = ""
     def cfStackName = ""
-		if(params.service_name != null && !params.service_name.isEmpty()){
-			serviceName = params.service_name
-			cfStackName = params.service_name
-		}
+	
+	if(params.service_name != null && !params.service_name.isEmpty()){
```
deepusundar: groogy has a simpler way of checking null or empty. we can just use  if(params.service_name){..}

suryajak: Cool, will use it for next set of cleanup

PR #116, opened 2018-01-29
Fixed s3 svg mime type

chimpuz:
### Requirements

### Benefits

- Fix the mime-type for the svg files uploaded.

### Possible Drawbacks

- None

### Applicable Issues

- Correct website rendering.


chimpuz:
Need to test with some build. Works fine internally.

suryajak:
Can you submit this PR against v1.3.1 or v1.4? Both are our vnext branches, and that way this will get validated as part of our daily build.

PR #132, opened 2018-02-07
Delete workflow uses SCM Module

suryajak:
### Description of the Change

Integrate scm module into delete flow

### Benefits

- modular code 
- separation of concerns 
- enable future scenarios to allow other SCM plugins

### Possible Drawbacks

possible that existing functionality could be broken (Validated on daily stack and it works)

### DIFF
--- a/delete-serverless-service-build-pack/Jenkinsfile
+++ b/delete-serverless-service-build-pack/Jenkinsfile
@@ -27,12 +27,6 @@ import groovy.transform.Field
 @Field def role = ''
 @Field def jq = '/usr/local/bin/jq'
 
-@Field def var_repo_core
-@Field def var_repo_loc
-
-@Field def var_bitbucket_base
-@Field def var_bitbucket_email
-
 node {
 	echo "Starting delete service job with params : $params"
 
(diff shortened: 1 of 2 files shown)


delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -1049,41 +882,16 @@ def disableCloudFrontDistribution(distributionID, _Etag, configFile, stage) {
...
-		error "deleteCloudFrontDistribution. "+e.getMessage()
-	}
-	*/
-}
-
-def loadServerlessConfig(String runtime, String configPackBBURL) {
+def loadServerlessConfig(String runtime) {
 
-	echo "loadServerlessConfig::" + runtime + "::" + configPackBBURL
+	def configPackURL = scmModule.getCoreRepoCloneUrl("serverless-config-pack")
```
deepusundar: Should this be pulled from from configuration?

suryajak: I want to consolidate all our platform specific stuff into one repo (jazz-build-module -> maybe rename this too a more meaningful name then). That will be for another story though. What I can do though is push this to a class member variable and reference it that way. Does that make sense?

PR #142, opened 2018-02-14
[WIP] US276433: Platform services parity with gitlab

SiniWilson:
### Requirements

Platform services parity with gitlab

### Description of the Change

Added webhook
Created new job
Added scm_type
Added SERVICE_TYPE

### Benefits

Platform services deployed as part of gitlab

### DIFF
--- a/gitlab-platform-services-build-pack/Jenkinsfile
+++ b/gitlab-platform-services-build-pack/Jenkinsfile
@@ -1,0 +1,65 @@
+#!groovy
+import groovy.json.JsonOutput
+import groovy.transform.Field
+
+// To be replaced as @Field def repo_credential_id = "value" for repo_credential_id, repo_base and repo_core
+@Field def repo_credential_id
+@Field def repo_base
+@Field def repo_core
+@Field def scm_type
+
+@Field def configModule
+@Field def configLoader
+@Field def jenkins_credential_id
+@Field def region
+
+node {
+  loadConfigModule(getBuildModuleUrl())
+
+  jenkins_credential_id = configLoader.JENKINS.CREDENTIAL_ID
+  region = configLoader.AWS.REGION
+
+withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: jenkins_credential_id, passwordVariable: 'PWD',
+usernameVariable: 'UNAME']]){
+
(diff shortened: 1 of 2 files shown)


gitlab-platform-services-build-pack/Jenkinsfile
```diff
@@ -0,0 +1,65 @@
...
+node {
+  loadConfigModule(getBuildModuleUrl())
+
+  jenkins_credential_id = configLoader.JENKINS.CREDENTIAL_ID
+  region = configLoader.AWS.REGION
+
+withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: jenkins_credential_id, passwordVariable: 'PWD',
+usernameVariable: 'UNAME']]){
+
+   def gitlab_repo_name  = env.gitlabSourceRepoName
```
suryajak: Why we are not using the same build-deploy-platform-services Jenkins file? You can use scm_type to extract repo and branch name specific to each SCM.

SiniWilson: There is already a story in future sprint to remove "build-deploy-platform-services" job with Suprita. 
What will be approach when Suprita removes that job. Can you please advise.

suryajak: build-deploy-platform-services will be merged into build packs so the logic will also move there.

SiniWilson: This job needs gitlab related plugins. And in the build-deploy-platform services does not have that one  and it contain bitbucket related plugins. I am not sure whether it will get any conflicts or we can have both the plugins.

suryajak: plugin configuration is during install time. We should configure the appropriate plugin for the job(s) depending on what was chosen. Writing similar business logic in multiple files is error prone, and has high cost of maintenance especially when the difference between the two is reading one set of env variables vs the other.

SiniWilson: I already approached the same way at first and I found some issues with that approach since it is xml. Thats why I opted theis approach. So as per my understanding we cant add the plugin configurations in xml as conditioanally. Also it needs some RND. It will be better if we in cooperate those in a separate story.  Could you please suggest.

PR #152, opened 2018-02-23
US276527 : Follow up PR - bug fixing

SiniWilson:
### Requirements

US276527 : Follow up PR - bug fixing

### Description of the Change

Bug fixing and modification for gitlab

### DIFF
--- a/jenkins-build-pack-api/Jenkinsfile
+++ b/jenkins-build-pack-api/Jenkinsfile
@@ -108,7 +108,7 @@ node() {
 		}
 		
 		if(configObj.service_id){	
-			config = loadServiceMetadata(configObj.service_id)
+			config = serviceConfigdata.loadServiceMetadata(configObj.service_id,configLoader)
 		}else{
 			error "Service Id is not available."
 		}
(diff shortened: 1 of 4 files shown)


jenkins-build-pack-api/Jenkinsfile
```diff
@@ -212,9 +203,9 @@ node() {
 					if(domain != "platform"){
```
suryajak: We don't need this if clause. Platform services endpoint info will be updated and so will be the API Docs.

jenkins-build-pack-api/Jenkinsfile
```diff
@@ -301,12 +292,12 @@ def buildLambda(runtime) {
 /**
 	Generate the swagger file specific to each environment
 */
-def generateSwaggerEnv(service, env, deploymentNode, apiHostName, domain, serviceName, config) {
-    sh "sed -i -- 's/{lambda_function_name}/" + serviceName + "-" + env + "/g' swagger/swagger.json"
+def generateSwaggerEnv(env, deploymentNode, apiHostName,config) {
+    sh "sed -i -- 's/{lambda_function_name}/${configLoader.INSTANCE_PREFIX}-${config['service']}-${env}/g' swagger/swagger.json"
```
suryajak: Thanks for taking the extra step and making the code uniform.

jenkins-build-pack-api/Jenkinsfile
SiniWilson: In cooperated the review comment. But that if clause is needed because , that block is for creating the subscription filters for the user services.

PR #155, opened 2018-02-23
US276527: bug fixing related to gitlab

SiniWilson:
### Requirements
US276527: bug fixing related to gitlab


jenkins-build-pack-api/Jenkinsfile
```diff
@@ -78,7 +78,7 @@ node() {
 	
     stage('Checkout code base') {
 		if (params.domain  && params.domain.trim() != "" ) {			
-			if(domain != "platform"){
+			if(domain != "platform" && configLoader.SCM.TYPE != "gitlab"){
```
suryajak: Why is this required? We ideally shouldn't have the diff between bitbucket and gitlab, so please comment above this line why is that required?

jenkins-build-pack-lambda/Jenkinsfile
```diff
@@ -73,7 +73,7 @@ node ()  {
 	def domain = params.domain
 
 	if(params.domain && params.domain.trim() != ""){
-		if(params.domain != "platform"){
+		if(params.domain != "platform" && configLoader.SCM.TYPE != "gitlab" ){
```
suryajak: Same as above - Why is this required?

jenkins-build-pack-website/Jenkinsfile
```diff
@@ -78,7 +78,9 @@ node ()  {
     stage('Checkout code base') {
 
 		if (domain && domain.trim()!="") {
-			repo_name = params.domain + "-" + service
+			if(configLoader.SCM.TYPE != "gitlab" ){
```
suryajak: Same as above - Why is this required?

jenkins-build-pack-api/Jenkinsfile
SiniWilson: This is because the diversity in the repo naming of platform services and user services.
In Gitlab the same webhook trigger job is using for platform services and user services.
So this if block is needed.

jenkins-build-pack-lambda/Jenkinsfile
SiniWilson: This is because the diversity in the repo naming of platform services and user services.
In Gitlab the same webhook trigger job is using for platform services and user services.
So this if block is needed.

jenkins-build-pack-website/Jenkinsfile
SiniWilson: This is because the diversity in the repo naming of platform services and user services.
In Gitlab the same webhook trigger job is using for platform services and user services.
So this if block is needed.

jenkins-build-pack-api/Jenkinsfile
suryajak: I am not sure I quite follow. I believe this is a bug that needs to be fixed elsewhere (the deviation between how gitlabs jobs are triggered vs bitbucket jobs are triggered). Can you point those differences?

PR #187, opened 2018-03-19
bugfix for services api

supritat:
### Requirements


jazz_services/index.js
```diff
@@ -133,10 +133,6 @@ module.exports.handler = (event, context, cb) => {
                     logger.error('Error occured. ' + JSON.stringify(error, null, 2));
```
suryajak: The line `var data = result.fetchServices;` needs to be after if(error) 

Also `if(error) {`
should have a `return handleResponse(error, data ....);`

supritat: changes are made as per the comment

PR #213, opened 2018-04-20
Follow up PR

SiniWilson:
### Requirements

### DIFF
--- a/jazz-build-module/environment-deployment-metadata-loader.groovy
+++ b/jazz-build-module/environment-deployment-metadata-loader.groovy
@@ -19,7 +19,6 @@ echo "Environment and deployment metadata module loaded successfully"
 @Field def g_service_branch 
 @Field def g_git_commit_hash
 @Field def g_git_repo_url
-@Field def g_git_token
 @Field def g_build_url
 @Field def g_build_id
 @Field def g_environment_logical_id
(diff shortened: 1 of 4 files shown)


jazz-build-module/environment-deployment-metadata-loader.groovy
```diff
@@ -76,10 +76,9 @@ def getRepoURL() {
 }
 
 def getEnvironmentLogicalId() {
-	if (g_environment_logical_id == null) {
-		def token = g_login_token
+	if (g_environment_logical_id == null && service_config['domain'] != "jazz") {
```
suryajak: Why are we having to exclude jazz namespace?

jazz-build-module/environment-deployment-metadata-loader.groovy
```diff
@@ -150,10 +149,9 @@ def createPromotedEnvironment(environment_logical_id, created_by) {
 def checkIfEnvironmentAvailable(environment_logical_id) {	
 	def isAvailable = false
 	try {
-		if (environment_logical_id) {
-			def token = g_login_token
+		if (environment_logical_id && service_config['domain'] != "jazz") {
```
suryajak: Does this mean we can't have non-prod environments for jazz namespace?

jazz-build-module/environment-deployment-metadata-loader.groovy
SiniWilson: While deploying jazz services , the environment handlers and apis also login may not deployed.

suryajak: ahh, good point. We need to find another way to work around that. Lets come up with alternative ways that are better than the current "if" clauses.

PR #233, opened 2018-04-27
Public endpoint in service onboarding parity with internal

SiniWilson:
### Requirements
Handling public endpoints

### DIFF
--- a/jazz_create-serverless-service/components/validate-arn.js
+++ b/jazz_create-serverless-service/components/validate-arn.js
@@ -1,0 +1,26 @@
+/** 
+  @module: validate-arn.js
+  @description: Validates ARN value for a lambda function in this case Authorizer
+  @author:
+  @version: 1.0
+**/
+
+var validateARN = (arn) => {
+
+  var arnvalues = arn.split(":");
+  var isarnvalid = true;
+  //validate if arn is a valid Lambda
+  if(arn.indexOf("arn:aws:lambda") !== 0)
+    isarnvalid = false;
+  if(arnvalues[5] !== "function")
+    isarnvalid = false;
+  if(arnvalues.length > 7)
+    isarnvalid = false;
+  return isarnvalid;
+
+};
+
+module.exports = (arn) => {
+    var isarnvalid = validateARN(arn);
+    return isarnvalid;
+};
--- a/jazz_create-serverless-service/components/validate-arn.js
+++ b/jazz_create-serverless-service/components/validate-arn.js
@@ -6,21 +6,21 @@
 **/
 
 var validateARN = (arn) => {
+  if (arn) {
+    var arnvalues = arn.split(":");
+    var isarnvalid = true;
+    //validate if arn is a valid Lambda
+    if (arn.indexOf("arn:aws:lambda") !== 0 || arnvalues[5] !== "function" || arnvalues.length > 7)
+      isarnvalid = false;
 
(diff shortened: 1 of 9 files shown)


jazz_create-serverless-service/components/validate-arn.js
```diff
@@ -0,0 +1,26 @@
...
+  @author:
+  @version: 1.0
+**/
+
+var validateARN = (arn) => {
+
+  var arnvalues = arn.split(":");
+  var isarnvalid = true;
+  //validate if arn is a valid Lambda
+  if(arn.indexOf("arn:aws:lambda") !== 0)
```
svidhani: - indentation "if ("
- why not use || and combine the if conditions

jazz_create-serverless-service/components/validate-arn.js
```diff
@@ -0,0 +1,26 @@
+/** 
+  @module: validate-arn.js
+  @description: Validates ARN value for a lambda function in this case Authorizer
+  @author:
+  @version: 1.0
+**/
+
+var validateARN = (arn) => {
+
+  var arnvalues = arn.split(":");
```
svidhani: what if arn is null or undefined. Have you already validated that? Even if you have for all public(exportable) functions you should check if its not null/undefined.
Please add a unit test for this

jazz_create-serverless-service/index.js
```diff
@@ -75,24 +76,24 @@ module.exports.handler = (event, context, cb) => {
...
-                            } else {
-                                logger.error("Updated service catalog with failed status.");
-                                return cb(JSON.stringify(errorHandler.throwInternalServerError(err.message)));
-                            }
-                        });
+                if (err.jenkins_api_failure) {
+                    serviceDataObject.body = {
+                        "STATUS": "creation_failed"
+                    };
+                    crud.update(serviceId, serviceDataObject, function (serviceUpdateError, results) {
```
svidhani: you can use  (serviceUpdateError, results) => here and other places in the code

jazz_create-serverless-service/index.js
```diff
@@ -221,9 +223,30 @@ module.exports.handler = (event, context, cb) => {
             if (event.body.slack_channel) {
                 inputs.SLACKCHANNEL = event.body.slack_channel;
             }
-            if ((event.body.service_type === "api" || event.body.service_type === "function") && (event.body.require_internal_access !== null)) {
+
+            if ((event.body.service_type === "api" || event.body.service_type === "function") && (event.body.require_internal_access)) {
```
svidhani: you can extract event.body into a variable and re-use it in your code

jazz_create-serverless-service/components/validate-arn.js
SiniWilson: Changed with OR

jazz_create-serverless-service/components/validate-arn.js
SiniWilson: Already validating the arn. Also included validation in the exportable funciton. Unit test case is handling in the separate story. So leaving the unit test case part as of now.

jazz_create-serverless-service/index.js
SiniWilson: changed

jazz_create-serverless-service/index.js
SiniWilson: In cooperated the change

PR #239, opened 2018-05-01
Delete service was using guid which is replaced by uuid.v4

svidhani:
### Requirements

Delete service was using the deprecated GUID package which is replace with uuid v4.

### Description of the Change
Updated the code and unit test 

### Benefits

Will not have to maintain a deprecated dependency and would enforce consistency in the code

### Possible Drawbacks

NA


suryajak:
@svidhani  Curious, how did the unit tests pass? They should have failed with a similar issue, no?

svidhani:
Verified on gitlab(stack) that this works. Also, updated unit test.

PR #262, opened 2018-05-09
V1.4.6

hks93:
Code refactorized in Request -Promise format
Unit tests updated


jazz_create-serverless-service/index.js
```diff
@@ -31,15 +31,17 @@ const crud = require("./components/crud")(); //Import the utils module.
     @author:
     @version: 1.0
 **/
-
-module.exports.handler = (event, context, cb) => {
+var user_id =  undefined;
```
svidhani: it is anyways undefined, no need to specify it

jazz_create-serverless-service/index.js
```diff
@@ -101,201 +103,208 @@ module.exports.handler = (event, context, cb) => {
...
+        }
 
+        if (service_creation_data.service_type === "website") {
+            var create_cloudfront_url = "true";
+            serviceMetadataObj.create_cloudfront_url = create_cloudfront_url;
+            inputs.RUNTIME = 'n/a';
+        }
+        // Add rate expression to the propertiesObject;
+        if (service_creation_data.service_type === "function") {
+            if (service_creation_data.rateExpression !== undefined) {
```
svidhani: if (service_creation_data.rateExpression) will do

jazz_create-serverless-service/package.json
```diff
@@ -6,7 +6,8 @@
   "private": true,
   "author": [],
   "scripts": {
-    "test": "mocha"
+    "test": "mocha",
+    "tst": "nyc --reporter=html --reporter=text mocha"
```
svidhani: did you mean to check this in - if yes then I would put a better command like test-report or test-coverage or something

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,623 @@ const awsContext = require('aws-lambda-mock-context');
...
+
+          return obj.callback(null, responseObject_serviceOnboarding, responseObject_serviceOnboarding.body);
+        }
+        //return obj.callback(null, responseObject, responseObject.body);
+      });
+
+      //trigger the spy wrapping the logger by calling handler() with valid params
+      let callFunction = index.handler(event, context, (err, res) => {
+        reqStub.restore()
+        if (err) {
```
svidhani: i would separate these tests. One for err and other for success. in unit test its better to have test do just one thing

svidhani: Please update in other tests as well. the idea is that unit test results should be predictable.

jazz_create-serverless-service/index.js
```diff
@@ -31,15 +31,17 @@ const crud = require("./components/crud")(); //Import the utils module.
...
+var serviceDataObject;
+var handler = (event, context, cb) => {
 
     var errorHandler = errorHandlerModule();
     var config = configObj(event);
     logger.init(event, context);
 
-    var serviceId;
-    var serviceDataObject;
+    serviceId;
```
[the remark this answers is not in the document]
hks93: Will Remove the 2 statements : 
Putting the variables as global because variables are used in the subsequent functions that are called in index. handler, due to which the lifetime of the variable is the entire lifetime of index handler - ie the entire lifetime of this API execution. Declaring them global as to avoid increasing the parameters passed of the subfunctions and keeping them as modular as possible

jazz_create-serverless-service/index.js
hks93: fixed

jazz_create-serverless-service/index.js
hks93: fixed

jazz_create-serverless-service/package.json
hks93: done

jazz_create-serverless-service/test/test.js
hks93: Added assert.fail() on error case: can you give more clarification on the exact expectation ...did not understand

svidhani: so the idea is that you should avoid doing if and else. for if(err) you should have a different test and assert that it will be an error and for success there should be a separate unit test where you assert success. you should setup your test for each kind of scenario and not have if and else.
does this help?

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+        }
+      };
+      event.stage = "dev";
+      let config = configObj(event);
+      // wrapping requests 
+      reqStub = sinon.stub(request, "Request", (obj) => {
+        // Matching response Object to the corresponding Request call
+        if (obj.uri === (config.SERVICE_API_URL + config.TOKEN_URL)) {
+          return obj.callback(null, responseObject_getToken, responseObject_getToken.body);
+        } else if (obj.uri === "https://{conf-apikey}.execute-api.{conf-region}.amazonaws.com/dev/jazz/services") {
```
svidhani: this is similar to what we discussed today. It should be predictable. I would split these into their own tests. This test is doing multiple things

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+
+          return obj.callback(null, responseObject_createService, responseObject_createService.body);
+
+        } else if (obj.url === '{conf-jenkins-host}/job/create-service/buildWithParameters') {
+
+          return obj.callback(null, responseObject_serviceOnboarding, responseObject_serviceOnboarding.body);
+        }
+        //return obj.callback(null, responseObject, responseObject.body);
```
svidhani: remove this and other commented code in other places

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+          "service_name": "test-service",
+          "service_type": "function",
+          "domain": "test-domain",
+          "runtime": "nodejs",
+          "approvers": ['tw1light_$pArkle'],
+          "rateExpression": "1 * * * ? *",
+          "slack_channel": "mlp_fim",
+          "require_internal_access": false,
+          "create_cloudfront_url": false //, //?
+          //"enableEventSchedule" : false
```
svidhani: remove

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+      let reqStub = sinon.stub(request, "Request", (obj) => {
+        return obj.callback(null, responseObject, responseObject.body);
+      });
+      index.getToken(config).then((res) => {
+        if (res && res === "ghd93-3240-2343") {
+          bool = true;
+        }
+        assert.isTrue(bool);
+      })
+      reqStub.restore();
```
svidhani: normally when you stub you should also verify sinon.assert.calledOnce(or twice) on the stub to ensure it is indeed being called.

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+          "service_name": "test-service",
+          "service_type": "function",
+          "domain": "test-domain",
+          "runtime": "nodejs",
+          "approvers": ['tw1light_$pArkle'],
+          "rateExpression": "1 * * * ? *",
+          "slack_channel": "mlp_fim",
+          "require_internal_access": false,
+          "create_cloudfront_url": false //, //?
+          //"enableEventSchedule" : false
```
svidhani: remove

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+      };
+    });
+    it("should should return error when passed an invalid rateExpression", () => {
+      let cronValues = [null, "", "P!nk!e_P!e"];
+      let authToken = "temp-auth-token";
+      let config = configObj(event);
+      let bool = false;
+      for (let cron in cronValues) {
+        service_creation_data.rateExpression = cron
+        index.getServiceData(service_creation_data, authToken, config).then().catch((errorMsg) => {
```
svidhani: remove then()

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+          }
+        }
+        assert.isTrue(bool);
+      })
+    });
+    it("should return input object with METADATA values for valid input parameters for service type function (event_source s3)", () => {
+      let authToken = "temp-auth-token";
+      let bool = false;
+      service_creation_data["enableEventSchedule"] = true;
+      //service_creation_data.event_source_ec2 = "temp-url";
```
svidhani: remove

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+    it("should return input object with METADATA values for valid input parameters for service type function (event_source e2c)", () => {
+      let authToken = "temp-auth-token";
+      let bool = false;
+      service_creation_data["enableEventSchedule"] = true;
+      service_creation_data.event_source_ec2 = "temp-url";
+      service_creation_data.event_action_ec2 = "temp-url";
+      service_creation_data.event_source_ec2 = "sample-test-data"
+      let config = configObj(event);
+      index.getServiceData(service_creation_data, authToken, config).then((input) => {
+        if (input.METADATA.eventScheduleEnable && input.METADATA.eventScheduleRate != null && input.METADATA.eventScheduleRate != "") {
```
svidhani: expectation should be to set correctly that you can assert and not have to do 'if' here (similar)

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+          "service_name": "test-service",
+          "service_type": "function",
+          "domain": "test-domain",
+          "runtime": "nodejs",
+          "approvers": ['tw1light_$pArkle'],
+          "rateExpression": "1 * * * ? *",
+          "slack_channel": "mlp_fim",
+          "require_internal_access": false,
+          "create_cloudfront_url": false //, //?
+          //"enableEventSchedule" : false
```
svidhani: remove

jazz_create-serverless-service/test/test.js
```diff
@@ -22,234 +22,667 @@ const awsContext = require('aws-lambda-mock-context');
...
+        assert.fail();
+      }).catch((err) => {
+        if (err.jenkins_api_failure && err.message != null) {
+          bool = true;
+        }
+        assert.isTrue(bool);
+      })
+      reqStub.restore();
+    })
+    it("should return error message when jenkins job has failed", () => {
```
svidhani: give extra lines between every 'it'. other place too.

jazz_create-serverless-service/index.js
```diff
@@ -101,201 +99,208 @@ module.exports.handler = (event, context, cb) => {
...
-                    }
-                });
-            } catch (e) {
-                logger.error('Error during startServiceOnboarding: ' + e.message);
-                reject(e);
-            }
-        });
-    }
+            userlist = approvers.reduce((stringSoFar, approver) => {
+                return stringSoFar + util.format("name=%s&", approver);
```
svidhani: do you want to return without resolve/reject?

jazz_create-serverless-service/test/test.js
hks93: this test is doing only one thing ... it is verifying that index handler returns success message when true case scenario is passed : 
since in index. handler makes chained .then() function call, I would have to mock request call from all subsequent functions >
Induvidual functions are tested in detial below this one

jazz_create-serverless-service/test/test.js
hks93: done

jazz_create-serverless-service/test/test.js
hks93: done

jazz_create-serverless-service/test/test.js
hks93: done

jazz_create-serverless-service/test/test.js
hks93: done

jazz_create-serverless-service/test/test.js
hks93: done

jazz_create-serverless-service/test/test.js
hks93: done

jazz_create-serverless-service/test/test.js
hks93: done

jazz_create-serverless-service/index.js
```diff
@@ -31,17 +31,15 @@ const crud = require("./components/crud")(); //Import the utils module.
     @author:
     @version: 1.0
 **/
-
-module.exports.handler = (event, context, cb) => {
+var user_id;
```
suryajak: You have changed the scope of user_id, serviceId, serviceDataObject, is that intentional?

jazz_create-serverless-service/test/test.js
hks93: the expectation is that metadata values should be populated by the function 
if input.METADATA values are not populated , the test case fails

jazz_create-serverless-service/test/test.js
svidhani: i am not talking abt the index handler but lines 256-266. these setups should be done one for each test.
Basically my question is why are you doing if/else here. the stub should return one uri and if there is possibility of multiple uri then it should be a separate test.

PR #264, opened 2018-05-09
[Feature] [Logs] All logs should be fetched according to the specified log level and all levels below that.

sidddutta:
### Requirements

### DIFF
--- a/jazz_logs/index.js
+++ b/jazz_logs/index.js
@@ -88,14 +88,27 @@ module.exports.handler = (event, context, cb) => {
 
 			querys.push(utils.setQuery("servicename", service));
 			querys.push(utils.setQuery("environment", env));
-			querys.push(utils.setQuery("log_level", logType));
 
 			//Query to filter Control messages
 			querys.push(utils.setQuery("!message", "START*"));
 			querys.push(utils.setQuery("!message", "END*"));
 			querys.push(utils.setQuery("!message", "REPORT*"));
 
-			logger.info("QueryObj: "+ JSON.stringify(querys));
+			if (logType) {
+				var log_type_config = [];
+				_.forEach(config.LOG_LEVELS, function(value, key) {
+					log_type_config.push(value.Type);
+				});
+				
+				if(_.includes(log_type_config, logType.toLowerCase())) {
+					querys.push(utils.setLogLevelQuery(config.LOG_LEVELS, "log_level", logType.toLowerCase()));
+				} else {
+					logger.info("Only following values are allowed for logger type - " + log_type_config.join(", "));
+					return cb(JSON.stringify(errorHandler.throwInputValidationError("Only following values are allowed for logger type - " + log_type_config.join(", "))));
+				}
+			}
+
+			logger.info("QueryObj: " + JSON.stringify(querys));
 
 			var servCategory = [];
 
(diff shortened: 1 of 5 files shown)


jazz_logs/index.js
```diff
@@ -88,14 +88,27 @@ module.exports.handler = (event, context, cb) => {
...
 
 			//Query to filter Control messages
 			querys.push(utils.setQuery("!message", "START*"));
 			querys.push(utils.setQuery("!message", "END*"));
 			querys.push(utils.setQuery("!message", "REPORT*"));
 
-			logger.info("QueryObj: "+ JSON.stringify(querys));
+			if (logType) {
+				var log_type_config = [];
+				_.forEach(config.LOG_LEVELS, function(value, key) {
```
svidhani: consider using ES5 native function. you don't need lodash here

jazz_logs/index.js
```diff
@@ -88,14 +88,27 @@ module.exports.handler = (event, context, cb) => {
...
+					log_type_config.push(value.Type);
+				});
+				
+				if(_.includes(log_type_config, logType.toLowerCase())) {
```
svidhani: includes is part of Array.prototype. please consider using that

jazz_logs/utils.js
```diff
@@ -49,6 +50,32 @@ var set_query = function (type, value){
 	return query;
 };
 
+var set_log_level_query = function (LOG_LEVEL_CONFIG, type, value){
+	var query =  {
+		"query_string": { 
+			"query": type + ":" + value
+		} 
+	};
+	var requestedLogType =  _.filter(LOG_LEVEL_CONFIG, function(configObject) {
```
svidhani: Array.prototype.filter can be used here

jazz_logs/index.js
sidddutta: We are using NodeJs v4.3. Array.prototype is not supported for v4.3. It is supported from NodeJs v6.x and above. 

Polyfill may be used but don't want to use that to just replace  _.include  method for now.

jazz_logs/index.js
sidddutta: Replaced Lodash forEach with native for loop.

jazz_logs/utils.js
sidddutta: Done

sidddutta:
@svidhani @suryajak 
PR comments are updated except the one. Please review and merge.

jazz_logs/index.js
```diff
@@ -88,14 +88,27 @@ module.exports.handler = (event, context, cb) => {
...
 
 			//Query to filter Control messages
 			querys.push(utils.setQuery("!message", "START*"));
 			querys.push(utils.setQuery("!message", "END*"));
 			querys.push(utils.setQuery("!message", "REPORT*"));
 
-			logger.info("QueryObj: "+ JSON.stringify(querys));
+			if (logType) {
+				var log_type_config = [];
+				for(var count =0; count < config.LOG_LEVELS.length; count++ ) {
```
svidhani: consider using a map here (better for performance in general)
eg:
let log_type_config = config.LOG_LEVELS.map(logLevel => logLevel.Type);
what do you think? you can do this in other place in your code.

sidddutta: @svidhani  This actually piqued my interest. For normal array iteration map may be 70%~80% faster compared to native for loop iteration.
https://jsperf.com/sidd-test-for-vs-map

Though, I think the performance with callback almost comes to same I guess.
I have changed in both the places.

PR #271, opened 2018-05-14
[WIP] multi environment delete flow

SiniWilson:
### Requirements

This PR will helps to complete the delete flow in multi environment

### DIFF
--- a/jazz_services-handler/components/crud/get.js
+++ b/jazz_services-handler/components/crud/get.js
@@ -1,0 +1,34 @@
+const request = require('request');
+
+module.exports = (inputs, callback) => {
+
+	var svcPayload = {
+		uri: inputs.SERVICE_API_URL + inputs.SERVICE_API_RESOURCE + "?domain=" + inputs.DOMAIN + "&service=" + inputs.SERVICE_NAME,
+		method: 'GET',
+		headers: {
+			'Authorization': inputs.TOKEN
+		},
+		rejectUnauthorized: false
+	};
+
+	request(svcPayload, function (error, response, body) {
+		if (response.statusCode === 200 && body) {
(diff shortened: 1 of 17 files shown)


jazz_services-handler/components/crud/get.js
```diff
@@ -0,0 +1,34 @@
...
+		method: 'GET',
+		headers: {
+			'Authorization': inputs.TOKEN
+		},
+		rejectUnauthorized: false
+	};
+
+	request(svcPayload, function (error, response, body) {
+		if (response.statusCode === 200 && body && body.data ) {
+			var results = JSON.parse(body);
```
svidhani: if body needs to be parsed then body.data in if condition will give you undefined.

jazz_services-handler/index.js
```diff
@@ -123,16 +125,26 @@ var processRecord = function (record, configData,authToken) {
...
-			logger.info("found " + payload.Item.EVENT_TYPE.S + " event with sequence number: " + sequenceNumber);
-			resolve({
-				"interested_event": true,
-				"payload": payload.Item
-			});
-		} 
+		var kinesisPayload = JSON.parse(new Buffer(encodedPayload, 'base64').toString('ascii'));
+		logger.info("kinesisPayload : " + JSON.stringify(kinesisPayload));
+		if (kinesisPayload.Item.EVENT_TYPE && kinesisPayload.Item.EVENT_TYPE.S) {
+			if (_.includes(configData.EVENTS.EVENT_TYPE, kinesisPayload.Item.EVENT_TYPE.S) &&
```
svidhani: you can avoid using includes here and use ES5 native includes. Also, other places

PR #283, opened 2018-05-17
Unit test for deployments API

supritat:
### Requirements


jazz_deployments/index.js
```diff
@@ -159,15 +159,13 @@ module.exports.handler = (event, context, cb) => {
...
-			} else if (error.result === "unauthorized") {
-				return cb(JSON.stringify(errorHandler.throwUnauthorizedError(error.message)));
 			} else {
 				return cb(JSON.stringify(errorHandler.throwInternalServerError("Unexpected error occurred.")))
 			}
 		});
 };
 
-function genericInputValidation(event) {
+var genericInputValidation = (event) => {
```
svidhani: why did you change all these to be function expression. you could have just exported functions

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1106 @@
...
+        }
+      }
+    };
+    AWS.mock("DynamoDB.DocumentClient", "query", (param, cb) => {
+      var dataObj = {
+        Items: [event.body]
+      }
+      return cb(null, dataObj);
+    });
+    reqStub = sinon.stub(request, "Request", (obj) => {
```
svidhani: please also verify that the stub was being called eg: sinon.assert.ccalledOnce(reqStub).
please update in other places as well where you are using stubs or spies

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1106 @@
...
+const request = require('request');
+const sinon = require('sinon');
+
 const index = require('../index');
+const logger = require("../components/logger.js");
+const configObj = require('../components/config.js');
+const crud = require('../components/crud')();
+const errorHandler = require("../components/error-handler.js")();
+
+describe('jazz_deployments', function () {
```
svidhani: you can have multiple "describes" one for each function so its easy to group those test suites to track them and update them. Having them in one describe is hard to identify what tests are testing which unit.

supritat: Worked as per comments.

jazz_deployments/test/test.js
supritat: Worked as per comments

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1181 @@
...
+      event.method = "POST";
+      event.path =  {};
+      message = '{"errorType":"InternalServerError","message":"unexpected error occurred"}';
+      AWS.mock("DynamoDB.DocumentClient", "put", (params, cb) => {
+        return cb(err, null);
+      });
+      index.handler(event, context,(err, res) => {
+        err.should.be.equal(message);
+        AWS.restore("DynamoDB.DocumentClient");
+        return res;
```
svidhani: why do you have to return here? remove return res from all the test where it is not used to assert later

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1181 @@
...
+        }
+      };
+      AWS.mock("DynamoDB.DocumentClient", "query", (param, cb) => {
+        var dataObj = {
+          Items: [event.body]
+        }
+        return cb(null, dataObj);
+      });
+      reqStub = sinon.stub(request, "Request", (obj) => {
+        console.log(obj);
```
svidhani: remove console everywhere

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1181 @@
...
+        if(obj.method.toLowerCase() === 'post') {
```
svidhani: I would split this into two tests and not do if/else. set it with post in one and get in another and make it predictable rather than using if/else.

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1181 @@
...
+        var dataObj = {
+          Items: [event.body]
+        }
+        return cb(null, dataObj);
+      });
+      var i=0;
+      reqStub = sinon.stub(request, "Request", (obj) => {
+        i++;
+        console.log(i)
+        if(obj.method.toLowerCase() === 'post') {
```
svidhani: same comment here. please don't do if/else. let the test do one thing only. please update in other tests as well

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1181 @@
...
+      });
+      index.handler(event, context, (err, res) => {
+        err.should.be.equal(message);
+        AWS.restore("DynamoDB");
+        return err;
+      })
+    });
+  
+    it("should successfully get deployment with provided path param using GET method", () => {
+      event.method = "GET";
```
svidhani: if you want you can add more describes here and make it simpler.
One each for post, get and delete. I see the setups are being repeated for each one of them and it can be done once(using beforeEach and afterEach) and group all these HTTP verbs in their own respective describes. What do you think?

jazz_deployments/index.js
supritat: Worked as per comments

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1181 @@
...
+        event.body.status = update_data.status;
+        return cb(null, dataObj);
+      });
+      AWS.mock("DynamoDB.DocumentClient", "update", (params, cb) => {
+        var dataObj = {
+          Items: [event.body]
+        };
+        return cb(null, dataObj);
+      });
+      var processDeploymentsUpdate = index.processDeploymentsUpdate(config, event.body, tableName, event.path.id);
```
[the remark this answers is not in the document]
supritat: Worked as per comments

jazz_deployments/test/test.js
supritat: Removed

jazz_deployments/test/test.js
supritat: Removed console

jazz_deployments/test/test.js
supritat: Worked as per comments

jazz_deployments/test/test.js
supritat: Worked as per comments

jazz_deployments/test/test.js
supritat: Worked as mentioned

jazz_deployments/index.js
```diff
@@ -544,4 +548,28 @@ function buildNowRequest(serviceDetails, config, refDeployment) {
 			reject("unable to find deployment details");
 		}
 	});
-}
\ No newline at end of file
+}
+
+const factory = {
```
suryajak: This variable name is confusing as it isn't a `factory` in traditional terms.

jazz_deployments/index.js
```diff
@@ -544,4 +548,28 @@ function buildNowRequest(serviceDetails, config, refDeployment) {
...
+	validateUpdateInput,
+	updateDeploymentDetails,
+	deleteServiceByID,
+	reBuildDeployment,
+	getToken,
+	getServiceDetails,
+	buildNowRequest
+}
+
+module.exports = factory;
```
suryajak: Looking at the rest of the code, looks like we can do away with this confusing variable `factory`. Can you remove it and just have that object be directly assigned here?

jazz_deployments/index.js
```diff
@@ -30,7 +30,7 @@ const crud = require("./components/crud")(); //Import the crud module.
 const request = require('request');
 const util = require('util');
 
-module.exports.handler = (event, context, cb) => {
+function handler (event, context, cb) {
```
suryajak: Can we just use the standard convention we have in the rest of our code?
`var x = function() { ...`

svidhani: This does not work when you have to stub it.

jazz_deployments/index.js
svidhani: You cannot return the functions directly in module.exports as that cannot be stubbed using sinon(in same module). The way to work around this is to pass an object that returns the functions. If naming is the issue we can come up with another name.

jazz_deployments/index.js
```diff
@@ -217,10 +214,10 @@ function genericInputValidation(event) {
 	});
 };
 
-function processDeploymentCreation(config, deployment_details, deploymentTableName) {
+function processDeploymentCreation (config, deployment_details, deploymentTableName) {
```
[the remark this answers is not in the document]
supritat: Sure

jazz_deployments/index.js
supritat: Variable name is replaced as mentioned. Second part will be addressed as part of another PR.

jazz_deployments/index.js
supritat: Done

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1275 @@
...
+    it("should indicate missing required data error while validating create payload for new deployment", () => {
+      event.query = {};
+      event.path = {};
+      event.body.service = "";
+      index.validateDeploymentDetails(config, event.body)
+        .catch(error => {
+          expect(error).to.include({
+            result: 'inputError'
+          });
+        });
```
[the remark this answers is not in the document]
suryajak: Similar test cases are missing for _list_ and _update_ scenarios

jazz_deployments/test/test.js
```diff
@@ -1,11 +1,1275 @@
...
+    });
```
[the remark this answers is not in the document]
suryajak: Similar test cases are missing for _list_ and _update_ scenarios

jazz_deployments/test/test.js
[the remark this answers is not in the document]
suryajak: @supritat:  Did you get a chance to look at this?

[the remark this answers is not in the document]
supritat: Latest merge includes the expected test cases

PR #303, opened 2018-05-28
:bug:: Fixing issues with the cloudfront for the bucket per service feature

SiniWilson:
### Requirements

This PR will helps to resolve the cloudfront access issue while deploying websites under bucket per service feature


rajeevr2715:
Seems fine.

PR #309, opened 2018-05-29
Assets api

supritat:
### Requirements


jazz_assets/components/crud/create.js
```diff
@@ -0,0 +1,66 @@
...
+    var params = {
+        Item: {
+            "id": assets_id,
+            "timestamp": timestamp
+			},
+        ReturnConsumedCapacity: "TOTAL",
+        TableName: global.ASSETS_TABLE
+    };
+
+    Object.keys(assets_data).map((key) => {
```
svidhani: although we have been mutating the array in the map but the right way to use the map is to return the value which create a new array. That is the proper usage of map without side effect. - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

jazz_assets/components/crud/update.js
```diff
@@ -0,0 +1,66 @@
...
+    };
+
+    var keys_list = global.global_config.ASSETS_EDITABLE_FIELDS;
+    var update_exp = "";
+    var attributeValues = {};
+    var attributeNames = {};
+    var count = 0;
+    keys_list.map((key) => {
+        var key_name = key;		
+		if(_.includes(_.keys(update_data), key)){
```
svidhani: you can replace lodash with ES5 native function here. You have done this before.

jazz_assets/components/utils.js
```diff
@@ -0,0 +1,118 @@
...
+		};
+	}
+	return filter_expression;
+};
+
+var toLowercase = (input_data) => {
+	var asset_data = {};
+	Object.keys(input_data).map((field) => {
+		if (input_data[field] && input_data[field].constructor !== Array) {
+			if (_.includes(global.global_config.CASE_SENSITIVE_FIELDS, field.toLowerCase())) {
```
svidhani: you can replace lodash here with ES5 native

jazz_assets/components/validation/common.js
```diff
@@ -0,0 +1,197 @@
...
+        });
+    }
+};
+var validateDataTypes = (field, prop_value) => {
+    var fields_type = global.global_config.FIELD_DATA_TYPES;
+    var field_status = false;
+    Object.keys(fields_type).map((type) => {
+        if (field === type) {
+            if (fields_type[type] === 'String') {
+                if (prop_value && (typeof prop_value == 'string' || prop_value instanceof String)) {
```
svidhani: "==="

jazz_assets/components/validation/common.js
```diff
@@ -0,0 +1,197 @@
...
+};
+
+var validateEnumValues = (assets_data, onComplete) => {
+    var invalid_fields = [];
+    Object.keys(assets_data).map((field) => {
+        if (assets_data[field]) {
+            var value = assets_data[field];
+            switch (field) {
+                case 'status':
+                    if (!_.includes(global.global_config.ASSET_STATUS, value)) {
```
svidhani: replace usage of lodash here and below

jazz_assets/index.js
```diff
@@ -0,0 +1,265 @@
...
+            }
+        });
+    });
+};
+
+function postSearch(assets_data) {
+    logger.debug("Inside postSearch");
+    return new Promise((resolve, reject) => {
+        crud.postSearch(assets_data, (error, data) => {
+            if(error) {
```
svidhani: indentation here and other places in the code =?> 'if ('

jazz_assets/components/utils.js
supritat: Using indexOf()

jazz_assets/components/validation/common.js
supritat: Done

jazz_assets/components/validation/common.js
supritat: Using indexOf()

jazz_assets/index.js
supritat: Done

jazz_assets/components/crud/create.js
supritat: Noted.

jazz_assets/config/dev-config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "ASSETS_TABLE": "{conf_stack_prefix}_Assets_Dev",
+    "DDB_REGION": "us-east-1"
```
suryajak: make this config driven

supritat: Done

jazz_assets/components/crud/postSearch.js
```diff
@@ -0,0 +1,77 @@
...
+     };
+
+     var keys_list = global.global_config.ASSET_SEARCH_OPTIONAL_FILTER_PARAMS;
+
+     // Generate filter string
+     keys_list.map((key) => {
+         var key_name = utils.getDatabaseKeyName(key);
+
+         if (query[key] && key_name) {
+             if (key_name === "type") {
```
svidhani: yes please remove and update to avoid any confusion

jazz_assets/components/crud/update.js
```diff
@@ -0,0 +1,68 @@
...
+            "id": assets_id
+        }
+    };
+
+    var keys_list = global.global_config.ASSETS_EDITABLE_FIELDS;
+    var update_exp = "";
+    var attributeValues = {};
+    var attributeNames = {};
+    var count = 0;
+    keys_list.map((key) => {
```
svidhani: ideally you should return the values from the map. I think i had mentioned this before with a link on the usage of the map.

jazz_assets/components/validation.js
```diff
@@ -0,0 +1,311 @@
...
+        }
+    });
+    return field_status;
+};
+
+function validateEnumValues(assets_data) {
+    logger.debug("Inside validateEnumValues");
+    return new Promise((resolve, reject) => {
+        var invalid_fields = [];
+        Object.keys(assets_data).map((field) => {
```
svidhani: same thing. return values which will be pushed to the array invalid_fields

jazz_assets/components/validation.js
```diff
@@ -0,0 +1,311 @@
...
+            });
+        }
+    });
+};
+
+function validateEditableFields(update_data, editableFields) {
+    logger.debug("Inside validateEditableFields");
+    var invalid_fields = _.difference(_.keys(update_data), _.values(editableFields));
+    invalid_fields.map((value) => {
+        delete update_data[value];
```
svidhani: are you sure you want to mutate the original object using map or do you want to use something like filter on the original object and extract what is needed into a new one and use it. Generally its not a good idea to mutate objects like that (specially using a map, its not functional).

jazz_assets/index.js
```diff
@@ -0,0 +1,292 @@
...
+global.global_config = global_config;
+
+function handler(event, context, cb) {
+    //Initializations
+    var errorHandler = errorHandlerModule();
+    logger.init(event, context);
+    logger.info(event);
+    var config = configObj(event);
+    global.config = config;
+    // global.global_config = global_config;
```
svidhani: remove?

supritat: Removed.

jazz_assets/components/validation.js
supritat: Using filter function to remove invalid fields.

jazz_assets/components/validation.js
supritat: Updated

jazz_assets/components/crud/update.js
supritat: Updated

jazz_assets/components/crud/postSearch.js
supritat: updated

jazz_assets/components/utils.js
```diff
@@ -0,0 +1,115 @@
...
+		return null
+	}
+};
+
+
+var createFilterExpression = (assets_data) => {
+
+	var asset_type = assets_data.asset_type;
+	var filter_expression = {};
+	if (asset_type === 's3' || asset_type === 'cloudfront' || asset_type === 'lambda') {
```
svidhani: General comment - we should start thinking of making the asset types config driven. If we keep on hardcoding these in our code then it will make it hard for us to add other services in future. Not for this PR but something to think about.

jazz_assets/components/validation.js
```diff
@@ -0,0 +1,313 @@
...
+            });
+        }
+    });
+};
+
+function validateEmptyFieldsVal(assets_data) {
+    logger.debug("Inside validateEmptyFieldsVal");
+    return new Promise((resolve, reject) => {
+        var invalid_fields = [];
+        Object.keys(assets_data).forEach((field) => {
```
svidhani: well replacing map with foreach is not what i meant and that is not always what you should do. i think i had suggested using filter may be. Just get the values from assets_Data where there is some value. What do you think?  
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

jazz_assets/components/validation.js
```diff
@@ -0,0 +1,313 @@
...
+        }
+    });
+    return field_status;
+};
+
+function validateEnumValues(assets_data) {
+    logger.debug("Inside validateEnumValues");
+    return new Promise((resolve, reject) => {
+        var invalid_fields = [];
+        Object.keys(assets_data).forEach((field) => {
```
svidhani: you could use filter here. example code:
assets_data.filter(field => {
  if(field === 'status' && global.global_config.ASSET_STATUS.includes(field)) {
     return field;
  }
  if(field === 'assets' && global.global_config.ASSET_STATUS.includes(field)) {
     return field;
  }

jazz_assets/test/test.js
```diff
@@ -0,0 +1,768 @@
+const chai = require('chai');
```
svidhani: header?

jazz_assets/test/test.js
```diff
@@ -0,0 +1,768 @@
...
+const assert = require('chai').assert;
+const expect = require('chai').expect;
+const should = require('chai').should();
+const chaiAsPromised = require('chai-as-promised');
+chai.use(chaiAsPromised);
```
svidhani: r u using all these?

jazz_assets/test/test.js
```diff
@@ -0,0 +1,768 @@
...
+          sinon.assert.calledOnce(genericInputValidation);
+        });
+        genericInputValidation.restore();
+      });
+
+      it("should indicate internal server error", () => {
+        var resObj = '{"errorType":"InternalServerError","message":"Unexpected Server Error"}';
+        const genericInputValidation = sinon.stub(index, "genericInputValidation");
+        index.handler(event, context,(error, data) => {
+          console.log(error)
```
svidhani: you probably want to remove this. Also in other places if any

jazz_assets/components/validation.js
supritat: Using filter

jazz_assets/components/validation.js
supritat: Done

jazz_assets/test/test.js
supritat: Included

jazz_assets/test/test.js
supritat: Removed unused module

jazz_assets/test/test.js
supritat: removed

jazz_assets/components/utils.js
supritat: Sure

jazz_assets/Jenkinsfile_Platform
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  stage('Triggering the Jenkins build') {
+    withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: config_loader.JENKINS.CREDENTIAL_ID, passwordVariable: 'PWD', usernameVariable: 'UNAME']]){
+    echo "Build triggered via branch::${env.BRANCH_NAME}"
+
+    def build_job = config_loader.API.API_BUILD_URI
+    def job_url = JenkinsLocationConfiguration.get().getUrl() + build_job
+
+    sh "curl -X GET -k -v -u \"$UNAME:$PWD\"  \"" + job_url + "&service_name=assets&domain=jazz&scm_branch=${env.BRANCH_NAME}\""
```
suryajak: By the time this v1.7.1 will start building, we will have a new way of triggering builds: https://github.com/tmobile/jazz/pull/327/files#diff-09b7261183a16494507309dd77e3bc5eR34. Can you incorporate the changes ahead of time?

jazz_assets/components/utils.js
```diff
@@ -52,7 +52,7 @@ var getDatabaseKeyName = (key) => {
 		"service": "service",
 		"domain": "domain",
 		"environment": "environment",
-		"type": "type",
+		"asset_type": "asset_type",
```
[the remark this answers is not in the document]
SupritaTotagi: There is no change needed in the installer side as it is replacing a column name.

jazz_assets/Jenkinsfile_Platform
supritat: Incorporated new changes as mentioned.

PR #313, opened 2018-05-30
[WIP] Ossmerge/asset handler

deepusundar:
### Requirements

Asset handler for handling events related to assets


suryajak:
I don't see any changes except for this folder created for the new service, so I am assuming you plan to update this PR with your changes. Can you update the title to have [WIP] as prefix?

For WIP PRs you don't want to approve until all the changes are done. As CRer's what you can do is check if the code is moving in the right direction and if not, have your comments in there so that we can do course correction early.

PR #317, opened 2018-05-31
 [WIP] codequality analysis

SiniWilson:
### Requirements


jazz-build-module/sonar-module.groovy
```diff
@@ -1,152 +1,118 @@
 #!groovy?
-import groovy.json.JsonSlurper
+import groovy.json.JsonSlurperClassic
 import groovy.json.JsonOutput
 import groovy.transform.Field
+import java.util.regex.*
```
suryajak: where do we use regex?

PR #327, opened 2018-06-05
Jenkins: deprecate credentials in favor of token

SiniWilson:
### Requirements

This PR will helps to get rid off from jenkins tokens for building jobs 

### Benefits

No need for jenkins tokens


jazz_scm-webhook/Jenkinsfile
```diff
@@ -1,52 +0,0 @@
-#!groovy
```
suryajak: Thanks for finding and removing dead files

suryajak:
The changes here look good. You still a PR on installer side for config values for build URIs. Once you submit that I will review them both.

PR #332, opened 2018-06-07
Fixing credential issue

SiniWilson:
### Requirements

### DIFF
--- a/jazz-build-module/sonar-module.groovy
+++ b/jazz-build-module/sonar-module.groovy
@@ -221,55 +221,58 @@ def resetConfig() {
 	}
 }
 
-def cleanupCodeQualityReports(){
-	try {
-		def cleanupList = getSonarProject()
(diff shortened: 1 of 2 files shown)


jazz-build-module/sonar-module.groovy
```diff
@@ -221,19 +221,26 @@ def resetConfig() {
 	}
 }
 
+
 def cleanupCodeQualityReports(){
+	withCredentials([[$class: 'UsernamePasswordMultiBinding', credentialsId: config_loader.CODE_QUALITY.SONAR.ADMIN_SONAR_CREDENTIAL_ID, passwordVariable: 'PWD', usernameVariable: 'UNAME']]) {
```
suryajak: Can you move withCredentials closer to where it is being used? So in this case move this inside deleteReports. On a similar note you don't need uname, pwd as arguments to deleteReports.

PR #351, opened 2018-06-14
Added sonar protocol and updated other config key names

svidhani:
### Requirements

### DIFF
--- a/jazz_codeq/config/dev-config.json
+++ b/jazz_codeq/config/dev-config.json
@@ -3,15 +3,15 @@
 	"SONAR_PASSWORD": "{sonar_creds}",
 	"SONAR_URL": "{sonar_url}",
 	"SONAR_PROJECT_KEY" : "{sonar_project_key}",
+	"SONAR_PROTOCOL": "http://",
 	"SONAR_ENV_SERVICE": "/api/measures/search_history",
-	"ALLOWED_METRICS": ["security", "code-coverage", "code-smells", "lines-of-code", "files", "vulnerabilities"],
+	"ALLOWED_METRICS": ["security", "code-coverage", "code-smells", "lines-of-code", "files"],
 	"METRIC_MAP": {
 		"security": "vulnerabilities",
 		"code-coverage": "coverage",
 		"code-smells": "code_smells",
 		"lines-of-code": "ncloc",
-		"files": "files",
-		"vulnerabilities": "high_severity_vulns"
+		"files": "files"
 	},
 	"REQUIRED_PARAMS": ["domain", "service", "environment"],
 	"TOKEN_URL": "/dev/jazz/login",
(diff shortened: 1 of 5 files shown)


jazz_codeq/config/dev-config.json
```diff
@@ -3,15 +3,15 @@
 	"SONAR_PASSWORD": "{sonar_creds}",
 	"SONAR_URL": "{sonar_url}",
```
suryajak: looks like this config value was incorrectly labeled, this should be SONAR_HOSTNAME and not URL. Can you fix it so that in code where we are joining this, it is not confusing?

jazz_codeq/utils.js
```diff
@@ -211,9 +211,9 @@ function getCodeqReport(metrics, branch, toDate, fromDate, query, config) {
 			metricString = metricString + config.METRIC_MAP[metrics[m]] + ",";
 		}
 
-		const component =  config.SONAR_PROJECT_KEY + "_" + query.domain + "_" + query.service + "_" + branch;
+		const component = config.SONAR_PROJECT_KEY + "_" + query.domain + "_" + query.service + "_" + branch;
 		const svcPayload = {
-			uri: config.SONAR_URL + config.SONAR_ENV_SERVICE + "?metrics=" + metricString + "&from=" + fromDate + "&to=" + toDate + "&component=" + component,
+			uri:  config.SONAR_PROTOCOL + config.SONAR_URL + config.SONAR_ENV_SERVICE + "?metrics=" + metricString + "&from=" + fromDate + "&to=" + toDate + "&component=" + component,
```
suryajak: this code will make sense once we rename sonar_url to sonar_hostname

jazz_codeq/config/dev-config.json
```diff
@@ -3,15 +3,15 @@
...
-	"ALLOWED_METRICS": ["security", "code-coverage", "code-smells", "lines-of-code", "files", "vulnerabilities"],
+	"ALLOWED_METRICS": ["security", "code-coverage", "code-smells", "lines-of-code", "files"],
 	"METRIC_MAP": {
 		"security": "vulnerabilities",
 		"code-coverage": "coverage",
 		"code-smells": "code_smells",
 		"lines-of-code": "ncloc",
-		"files": "files",
-		"vulnerabilities": "high_severity_vulns"
+		"files": "files"
```
suryajak: also what is the deal with this metric `vulnerabilities`? @rajeevr2715 can you confirm that owasp is running in gitlab stack and is publishing data to sonar?

rajeevr2715: @suryajak owasp is running perfectly, but "Dependency-Check" plugin was not installed in the SonarQube, so the owasp reports were not published into SonarQube. I have updated the container script and installed the plugin.

PR #352, opened 2018-06-14
Code reorganization to speed up install time

SiniWilson:
### Requirements

This PR will helps to reorganize jazz core that will help to speed up the install time


core/jazz_delete-serverless-service/config/stg-config.json
```diff
@@ -1,5 +1,5 @@
 {
-    "DELETE_SERVICE_JOB_URL": "{conf-jenkins-host}job/delete-service/buildWithParameters",
+    "DELETE_SERVICE_JOB_URL": "{conf-jenkins-host}/job/delete-service/buildWithParameters",
```
suryajak: why is this changing as part of this PR?

builds/jazz-build-module/service-configuration-data-loader.groovy
```diff
@@ -278,6 +278,15 @@ def loadServiceConfigurationData() {
 			sh "sed -i -- 's/{conf-region}/${region}/g' ./config/dev-config.json"
 			sh "sed -i -- 's/{conf-region}/${region}/g' ./config/stg-config.json"
 			sh "sed -i -- 's/{conf-region}/${region}/g' ./config/prod-config.json"
+
```
suryajak: why is this change as part of this PR?

SiniWilson: I saw the jazz admin name of the jazz_services is written from the installer. As we are maintaining the service configuration during the deployment of the jazz core services, I think we dont want that part in installer. Thats why I had removed that from installer. So that we can reduce the jazz core dependency.

PR #364, opened 2018-06-19
Code reorganization to speed up install time

SiniWilson:
### Requirements

This PR will reorganize the jazz core so that we can improve the install time.


SiniWilson:
Need to rebase

PR #373, opened 2018-06-20
Fixed the bug for returning help on a selected metric

svidhani:
### Requirements


jazz_codeq/utils.js
```diff
@@ -109,6 +109,8 @@ function getMetrics(query, config, messages) {
 				let message = messages.MISSING_METRICS + invalid_metrics.join(", ");
 				result = { "error": message, "metrics": [] };
 				return result;
+			} else {
+				result = { "metrics": metrics };
```
suryajak: 👍

PR #375, opened 2018-06-21
Config objects should be local to the current execution 

suryajak:
### Description of the Change

Current behavior of our config object is that when a function executes and reads the config value, things work as expected. If the function now starts updating the "local" config object (add new property/update existing property), the current function works as expected. Now if another function gets run in the same container (parallel execution or execution while the function is still "warm") the subsequent execution will start seeing the updates (from the prior execution's updates). That leaves all Jazz services (core as well as user services) with a whole world of unintended consequences . 

The change here is to stay away from "require" (you get caching and shallow copy of objects) and use the native way of loading config files, where the references stay local and aren't influenced by other executions. 
 
### Benefits
NodeJS based functions can use config files w/o modifying the config object for other executions. 

### Possible Drawbacks
None


jazz_cloud-logs-streamer/components/config.js
```diff
@@ -14,30 +14,47 @@
...
-        // cw events default to dev
-        stage = 'dev';
-    } else {
-        stage = event.stage
-    }
+/**
+	Nodejs Template Project
+  @module: config.js
+  @description: Defines variables/functions to retrieve environment related data
+	@author:
```
devsatishm: Looks like some spacing issue in few files.

suryajak:
All jazz services along with templates are now updated to use the new config module

suryajak:
Unit tests for core/jazz_environment-event-handler/ are setup differently than most of the other unit tests. This update to config module was requiring more comprehensive updates to unit test setup as well as the unit tests for environment-event-handler and while I tried to update them, this was going way beyond the scope of this PR. We can have a different PR to make unit tests setup for environment-event-handler similar to the rest of them and subsequent to that we can have update to config module.

PR #401, opened 2018-06-27
Code Quality Graph Sizing

MichaelChinn:
code quality graph fits on screen with tiles when user enter page
some additional empty value handling

### DIFF
--- a/core/jazz-ui/src/app/pages/environment-codequality/env-codequality-section.component.scss
+++ b/core/jazz-ui/src/app/pages/environment-codequality/env-codequality-section.component.scss
@@ -67,7 +67,7 @@
 .section {
   position: relative;
   padding: 1rem 3rem;
-  height: calc(100vh - 100px);
+  height: calc(100vh - (13.15rem + 94px));
   background: white;
   box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.11);
   display: flex;
@@ -124,7 +124,8 @@
     background-color: #fdfdfd;
     border: solid 1px #dddddd;
     width: 135px;
-    display: inline-block;
+    display: inline-flex;
+    flex-direction: column;
     padding: 0.5rem;
     margin: -1px 0.5rem;
     text-align: center;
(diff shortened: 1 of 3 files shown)


core/jazz-ui/src/app/pages/environment-codequality/env-codequality-section.component.ts
```diff
@@ -172,19 +171,15 @@ export class EnvCodequalitySectionComponent implements OnInit {
...
-      .map((point) => {return point.y})
-      .reduce((a, b) => {
-      return Math.max(a, b);
-    }));
-    filterData.yMin = .9 * (data
-      .map((point) => {return point.y})
-      .reduce((a, b) => {
-      return Math.min(a, b);
-    }));
+    filterData.yMax = values.length ? 1.1 * (values
```
svidhani: do you think these values should be config driven?

core/jazz-ui/src/app/pages/environment-codequality/env-codequality-section.component.scss
```diff
@@ -144,6 +145,14 @@
     &:last-child {
       margin-right: 0;
     }
+
+    .metrics-card-content {
```
suryajak: Are we re-purposing the styles for metrics in code quality? Given they are two different components, perhaps the styles should be specific to them, no?

MichaelChinn: .metric-card refers to the tiles that go under the code quality graph since the code quality query returns {data: {metrics:[]}}. These styles will likely be copied into the metrics component and will sit as duplicates until I can migrate to a unified styling convention.

core/jazz-ui/src/app/pages/environment-codequality/env-codequality-section.component.ts
MichaelChinn: Yes, I think the best solution would be to use values send back from the api when you query the data. For now surya and I have agreed to use a 90-110% solution.

PR #443, opened 2018-07-20
Incorporating unit tests for environment handler

SiniWilson:
### Requirements

### DIFF
--- a/core/jazz_environment-event-handler/test/test.js
+++ b/core/jazz_environment-event-handler/test/test.js
@@ -881,12 +881,8 @@ describe('handler', () => {
     });
 
     index.handler(event, context, (err, res) => {
-      if (err) {
-        return err;
-      } else {
-        res.should.have.property('processed_events');
-        return res;
-      }
+      res.should.have.property('processed_events');
+      return res;
     });
 
     authStub.restore();
@@ -906,12 +902,8 @@ describe('handler', () => {
     });
     let message = 'User is not authorized to access this service';
     index.handler(event, context, (err, res) => {
-      if (err) {
-        return err;
-      } else {
-        res.should.have.property('failed_events');
-        return res;
-      }
+      res.should.have.property('failed_events');
+      return res;
     });
 
     authStub.restore();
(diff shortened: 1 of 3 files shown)


core/jazz_environment-event-handler/test/test.js
```diff
@@ -15,642 +15,1496 @@
...
+      body: { "message": "unautho" }
+    };
+
+    let authStub = sinon.stub(rp, 'Request').returns(Promise.resolve(responseObject));
+    requestPromiseStub.callsFake((obj) => {
+      return obj.callback(null, responseObject, responseObject.body);
+    });
+    let message = 'User is not authorized to access this service';
+    index.handler(event, context, (err, res) => {
+      if (err) {
```
svidhani: try not to use if and else in tests. Ensure your test is set up to either get err or data. You test should be deterministic. Split this and make 2 tests. One which surely gives and error and another with data(success). Please update other tests if you have taken similar approach

core/jazz_environment-event-handler/test/test.js
```diff
@@ -15,642 +15,1496 @@
...
+    };
+    let service_responseObject = {
+      statusCode: 200,
+      body: JSON.stringify(body)
+    };
+    processRequestStub.resolves(service_responseObject.body);
+    getServiceDetailsStub.resolves(service_responseObject.body);
+    requestPromiseStub.callsFake((obj) => {
+      if (obj.uri == "https://{conf-apikey}.execute-api.{conf-region}.amazonaws.com/prod/jazz/services?service=test-env-oss-3&domain=jazztesting&isAdmin=true") {
+        return obj.callback(null, service_responseObject, service_responseObject.bod
```
svidhani: new line

core/jazz_environment-event-handler/test/test.js
SiniWilson: Done

core/jazz_environment-event-handler/test/test.js
SiniWilson: Done

PR #444, opened 2018-07-21
Enable Dynamic API Ids processing using config and environments table 

suryajak:
### Description of the Change

Enable Dynamic API Ids processing using config and environments table. 

### Benefits

This extends the functionality of dynamic API Ids where Jazz Admins don't have to be constrained to the config values. Updates to API Ids for specific domain and/service can be made (addition or removal) without affecting existing API's as values will be preserved for each environment.

### Possible Drawbacks

None


builds/jazz-build-module/environment-deployment-metadata-loader.groovy
```diff
@@ -67,6 +67,29 @@ def getEnvironmentLogicalId() {
 	return g_environment_logical_id
 }
 
+def getEnvironmentInfo() {
+	if (g_service_config['domain'] != "jazz") {
+		def API_ENVIRONMENT_QUERY_URL = "${g_environment_api}?service=${g_service_config['service']}&domain=${g_service_config['domain']}"
+		def getEnvironments = sh(script: "curl -H \"Content-type: application/json\" -H \"Authorization:\"$g_login_token -X GET \"${g_environment_api}?service=${g_service_config['service']}&domain=${g_service_config['domain']}\" ", returnStdout: true).trim()
```
SiniWilson: We can use API_ENVIRONMENT_QUERY_URL

suryajak: Fixed it. It was copy paste code from another place which had similar issue, so fixed it there too.

PR #456, opened 2018-07-26
:bug: Fixing stack issue

SiniWilson:
### Requirements

### DIFF
--- a/core/jazz_usermanagement/config/dev-config.json
+++ b/core/jazz_usermanagement/config/dev-config.json
@@ -1,23 +1,29 @@
 {
-  "USER_POOL_ID":"{user_pool_id}",
+  "USER_POOL_ID": "{user_pool_id}",
   "REGION": "{region}",
   "USER_CLIENT_ID": "{user_client_id}",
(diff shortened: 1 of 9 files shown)


core/jazz_usermanagement/config/dev-config.json
```diff
@@ -1,23 +1,29 @@
...
+    "userpassword",
+    "usercode"
+  ],
   "SCM_TYPE": "{scm_type}",
-  "SCM_CONFIG" : {
-    "bitbucket" : {
-      "HOSTNAME": "{bb_service_host}",
+  "SCM_CONFIG": {
+    "bitbucket": {
+      "BASE_URL": "{base-url}",
```
suryajak: maybe call the replacement key as `scm_base_url`

suryajak: The key name was correct earlier as the value is a HOSTNAME and not a URL

SiniWilson: Done

PR #477, opened 2018-08-02
Support SQS as a trigger for functions

SiniWilson:
### Requirements

### DIFF
--- a/builds/jazz-build-module/aws-lambda-events-module.groovy
+++ b/builds/jazz-build-module/aws-lambda-events-module.groovy
@@ -25,14 +25,14 @@ def checkSQSAndAddLambdaTrigger(queueName, lambdaARN) {
 }
 
 def addLambdaTriggerToSqsQueue(isExists, queue_name, lambdaARN){
-    if(isExists){
-      def isDefined = listEventSourceMapping(queue_name, lambdaARN)
-      if(!isDefined){
-        createEventSourceMapping(queue_name, lambdaARN)
-      }
-    }else{
-      createSqsQueue(queue_name, lambdaARN)
+  if(isExists){
+    def isDefined = listEventSourceMapping(queue_name, lambdaARN)
+    if(!isDefined){
+      createEventSourceMapping(queue_name, lambdaARN)
     }
+  }else{
+    createSqsQueue(queue_name, lambdaARN)
+  }
 }
 
 def createSqsQueue(queue_name, lambdaARN){
--- a/builds/jazz-build-module/aws-lambda-events-module.groovy
+++ b/builds/jazz-build-module/aws-lambda-events-module.groovy
@@ -8,6 +8,7 @@ import groovy.transform.Field
 */
 
 @Field def config_loader
+@Field def queue_visibility_timeout = 165
 
 def initialize(configLoader){
   config_loader = configLoader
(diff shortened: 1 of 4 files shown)


builds/jenkins-build-pack-lambda/Jenkinsfile
```diff
@@ -299,6 +305,9 @@ def addEvents(def isScheduleEnabled, def isEc2EventEnabled, def isS3EventEnabled
 	if (!isS3EventEnabled) {
 		sedCommand = sedCommand + "; /#Start:isS3EventEnabled/,/#End:isS3EventEnabled/d"
 	}
+  if (!isSQSEventEnabled) {
+		sedCommand = sedCommand + "; /#Start:isSQSEventEnabled/,/#End:isSQSEventEnabled/d"
```
svidhani: shouldn't you have this in updateServerlessSQSEvent as well

builds/jazz-build-module/aws-lambda-events-module.groovy
```diff
@@ -0,0 +1,87 @@
...
+    sh "aws sqs get-queue-url --queue-name $queueName --profile cloud-api --output json"
+    echo "Queue exists and have access"
+    addLambdaTriggerToSqsQueue(true, queueName, lambdaARN)
+  } catch (ex) {
+    addLambdaTriggerToSqsQueue(false, queueName, lambdaARN)
+  }
+}
+
+def addLambdaTriggerToSqsQueue(isExists, queue_name, lambdaARN){
+    if(isExists){
```
svidhani: indentation here and other places

builds/jazz-build-module/aws-lambda-events-module.groovy
```diff
@@ -0,0 +1,87 @@
...
+        createEventSourceMapping(queue_name, lambdaARN)
+      }
+    }else{
+      createSqsQueue(queue_name, lambdaARN)
+    }
+}
+
+def createSqsQueue(queue_name, lambdaARN){
+  try{
+    sh "aws sqs create-queue --queue-name ${queue_name} --attributes '{\"VisibilityTimeout\": \"165\"}' --profile cloud-api --output json"
```
svidhani: why are you providing the timeout here?

SiniWilson: I am able to get the multiple lambda trigger working. Also the for the above question, by default the timeout period is 30. To increase that timeout period we have to provide the timeout period explicitly

builds/jazz-build-module/aws-lambda-events-module.groovy
SiniWilson: done

builds/jazz-build-module/aws-lambda-events-module.groovy
svidhani: but how did you come up with that number. @devsatishm are we providing a visibility timeout?

SiniWilson:
@svidhani  @devsatishm @suryajak  If we proceed with default timeout  of 30s we will get an error like 
**com.amazonaws.services.lambda.model.InvalidParameterValueException: Queue visibility timeout: 30 seconds is less than Function timeout: 160 seconds** . That is the reason I gave 165s as queue visibbility timeout.

builds/jazz-build-module/aws-lambda-events-module.groovy
```diff
@@ -8,12 +8,73 @@ import groovy.transform.Field
...
 
 def initialize(configLoader){
   config_loader = configLoader
 }
 
+def checkSQSAndAddLambdaTrigger(queueName, lambdaARN) {
+  try {
+    sh "aws sqs get-queue-url --queue-name $queueName --profile cloud-api --output json"
+    echo "Queue exists and have access"
+    addLambdaTriggerToSqsQueue(true, queueName, lambdaARN)
```
deepusundar: This is actually a tricky flow logic and can go wrong.  Since the flow logic is relied on an exception and it can happen for various reasons.  Best way would be,  parse the response and validate the proceed next.  Also for catch logic,  see if it failed for correct reason, validate the error message (since flow here proceeds to create an SQS queue

PR #489, opened 2018-08-08
[WIP] [apigee plug-in] Jenkins pipeline modifications to support apigee integration

michmerr:
### Description of the Change

Adds conditional support for the use of Apigee as an API gateway as an alternative to the AWS API Gateway. Specifically, these changes add the Jenkins pipeline code to create and delete services that use the Apigee API proxy. These require inputs from related UI modifications to define which API Gateway should be used.

The conditional logic is in the existing Jazz code, while most of the Apigee-specific code is in the apigee-module.groovy under jazz-build-module. Supporting files, including templates used in the proxy generation process are organized under the jazz-build-module/apigee directory.

### Benefits

Supports the feature that allow users to deploy services that use an Apigee API proxy.

### Possible Drawbacks

Additional configuration requirements.

### Applicable Issues

Existing deployments may require an update to indicate to which platform they are deployed.


devsatishm:
Changes are merged into new branch: apigee-extension [PR: https://github.com/tmobile/jazz/pull/490]

This PR can be closed now.

PR #493, opened 2018-08-09
Update delete service unit test configurations

hks93:


core/jazz_delete-serverless-service/components/config.js
```diff
@@ -27,26 +27,26 @@ const path = require('path');
...
       stage = 'prod';
-    }  
-  } 
+    }
+  }
 
   if (stage) {
-    var configFile = path.join(__dirname, `../config/${stage}-config.json`);
-
+    var core_config = ['dev','stg','prod'];
```
svidhani: why are you making this change?

hks93: The test cases for this story are designed on a high-level and not on a functional level.
All test cases call index.handler with  respective changes in the output 
the index file exports only one function ie) handler 
it is only possible to send 3 parameters to this function without modifying the index.js file 
since config is called from within the index file 
The values for invalidTest_config.json and validTest_config.json cannot be passed from outside 

The previous approach followed was that this file was loaded directly from config folder from the config.js file 
I have moved the test config files to test folder and added a condition in the config.js file  to fetch them from test-folder in case stage is not [dev,stg, prod] 

If these files have to be loaded from the test.js file itself 
the index should be refactorized 
And unit test cases should be rewritten - for this  the story has to be reestimated 

(Please suggest any other method if available )

suryajak:
I don't see the point of this PR unless we rewire the unit tests to address the actual issue.

PR #508, opened 2018-08-16
Fixing ui build issue with webhook

SiniWilson:
### Requirements

Jazz UI doesn't build when changes are pushed


core/jazz_environment-event-handler/index.js
```diff
@@ -491,29 +499,42 @@ function getServiceDetails(eventPayload, configData, authToken) {
...
 }
 
 var triggerBuildJob = function (result, payload, configData) {
   return new Promise((resolve, reject) => {
-    var output = JSON.parse(result);
-    if (!output.data && !output.data.services && output.data.services.length > 0) {
-      logger.error("Service details not found in service catalog: " + JSON.stringify(output));
-      var error = handleError(failureCodes.PR_ERROR_5.code, failureCodes.PR_ERROR_5.message);
-      return reject(error);
+    var output
```
suryajak: missing semi-colon here and other places between 518 - 525

core/jazz_environment-event-handler/index.js
```diff
@@ -263,18 +263,22 @@ var processEventInitialCommit = function (environmentPayload, configData, authTo
...
-          logger.error(`Error creating ${env} environment in catalog: ${JSON.stringify(response)}`);
-          return reject({
-            "error": `Error creating ${env} environment for ${environmentPayload.domain} "_" ${environmentPayload.service} in catalog`,
-            "details": response.body.message
-          });
-        }
-      });
+      if (environmentPayload.service === 'ui' && environmentPayload.domain === 'jazz') {
+        resolve();
+      } else {
```
suryajak: To avoid nested if else loops, I would just say `return resolve();`  and then remove the else altogether.

core/jazz_environment-event-handler/index.js
SiniWilson: done

core/jazz_environment-event-handler/index.js
SiniWilson: done

PR #517, opened 2018-08-20
Support S3 as event source for functions

SiniWilson:
### Requirements

User should be able to select S3 as function triggers.


svidhani:
I have not reviewed this. Please do not merge PRs until owners have reviewed it.

PR #518, opened 2018-08-20
Assets/pagination

deepusundar:
### Requirements

### DIFF
--- a/core/jazz_metrics/index.js
+++ b/core/jazz_metrics/index.js
@@ -132,7 +132,7 @@ function getToken(config) {
 function getAssetsDetails(config, eventBody, authToken) {
   return new Promise((resolve, reject) => {
     var asset_api_options = {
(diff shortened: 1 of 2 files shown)


core/jazz_metrics/index.js
```diff
@@ -132,7 +132,7 @@ function getToken(config) {
 function getAssetsDetails(config, eventBody, authToken) {
   return new Promise((resolve, reject) => {
     var asset_api_options = {
-      url: config.SERVICE_API_URL + config.ASSETS_URL + "?domain=" + eventBody.domain + "&service=" + eventBody.service + "&environment=" + eventBody.environment,
+      url: config.SERVICE_API_URL + config.ASSETS_URL + "?domain=" + eventBody.domain + "&service=" + eventBody.service + "&environment=" + eventBody.environment + "&limit=-1",
```
svidhani: why is this being done?

deepusundar: This fix is to pull all the assets for a service and environment.  The previous version assumed that a service won't be having too many assets created and ideally it will be less than max limit set

PR #520, opened 2018-08-20
Fix the deployment targets config update

svidhani:
Update the deployment targets in the config by splitting for each service type which will be updated by the config loader.

### DIFF
--- a/builds/jazz-build-module/service-configuration-data-loader.groovy
+++ b/builds/jazz-build-module/service-configuration-data-loader.groovy
@@ -344,10 +344,20 @@ def loadServiceConfigurationData() {
 			sh "sed -i -- 's/{jazz_admin_creds}/${config_loader.JAZZ.PASSWD}/g' ./config/stg-config.json"
 			sh "sed -i -- 's/{jazz_admin_creds}/${config_loader.JAZZ.PASSWD}/g' ./config/prod-config.json"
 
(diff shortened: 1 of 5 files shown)


builds/jazz-build-module/service-configuration-data-loader.groovy
```diff
@@ -344,10 +344,38 @@ def loadServiceConfigurationData() {
...
 			sh "sed -i -- 's/{jazz_admin_creds}/${config_loader.JAZZ.PASSWD}/g' ./config/prod-config.json"
 
-      sh "sed -i -- 's/{conf_deployment_targets}/${config_loader.JAZZ.DEPLOYMENT_TARGETS}/g' ./config/dev-config.json"
-      sh "sed -i -- 's/{conf_deployment_targets}/${config_loader.JAZZ.DEPLOYMENT_TARGETS}/g' ./config/stg-config.json"
-      sh "sed -i -- 's/{conf_deployment_targets}/${config_loader.JAZZ.DEPLOYMENT_TARGETS}/g' ./config/prod-config.json"
-      sh "sed -i -- 's/{conf_deployment_targets}/${config_loader.JAZZ.DEPLOYMENT_TARGETS}/g' ./config/test-config.json"
+      def apiOptions=""
+      def functionOptions=""
+      def websiteOptions=""
+      for (String item: config_loader.JAZZ.DEPLOYMENT_TARGETS.API) {
```
svidhani: had to do this as we want the array/list of options to be added and sed -i 's/ does only string replacement and we also want to retain the quotes. If there is any better way to do it I am all ears

PR #524, opened 2018-08-22
Hide Aws events panel

MichaelChinn:
Remove aws events panel for release

### DIFF
--- a/core/jazz_ui/src/app/secondary-components/create-service/internal/create-service.component.html
+++ b/core/jazz_ui/src/app/secondary-components/create-service/internal/create-service.component.html
@@ -483,7 +483,7 @@
                 </div>
             </section>
             <!-- start of aws events -->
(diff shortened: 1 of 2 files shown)


core/jazz_ui/src/app/secondary-components/create-service/internal/create-service.component.html
```diff
@@ -483,7 +483,7 @@ <h4 class="bold-title">
                 </div>
             </section>
             <!-- start of aws events -->
-            <section class="each-step-wrap aws-events" *ngIf="typeOfService == 'function'">
+            <section class="each-step-wrap aws-events" *ngIf="typeOfService == 'function' && false">
```
suryajak: In internal it should be true though, no?

MichaelChinn: Currently none of the files marked as internal do anything. Fixing that is part of the reverse pipeline story. This was just to keep them in sync.

PR #536, opened 2018-08-29
:bug: Fixing issues of duplicate key mapping in serverless yml

SiniWilson:
### Requirements

This PR will help to resolve the issues of duplicate key mapping in serverless.yml file while integrating kinesis, sqs, and dynamodb events

### Description of the Change

Modified the serverless.yml files 
Modified the build-pack-api
Modified the build-pack-lambda

### Benefits

Successful build for event subscribed services and not subscribed services.

### DIFF
--- a/builds/jenkins-build-pack-api/Jenkinsfile
+++ b/builds/jenkins-build-pack-api/Jenkinsfile
@@ -531,6 +531,9 @@ def writeServerlessFile(config, environment_logical_id){
 	sh "sed -i -- 's/\${file(deployment-env.yml):providerTimeout}/${config['providerTimeout']}/g' serverless.yml"
 	sh "sed -i -- 's/\${file(deployment-env.yml):securityGroupIds}/${config['securityGroupIds']}/g' serverless.yml"
 	sh "sed -i -- 's/\${file(deployment-env.yml):subnetIds}/${config['subnetIds'] }/g' serverless.yml"
+  sh "sed -i -- 's|\${file(deployment-env.yml):eventScheduleRate}|${config['eventScheduleRate']}|g' serverless.yml"
+	sh "sed -i -- 's/\${file(deployment-env.yml):eventScheduleEnable}/${config['eventScheduleEnable']}/g' serverless.yml"
+  sh "sed -i -- '/#Start:resources/,/#End:resources/d' ./serverless.yml"
 
 	if (config['artifact']) {
 		sh "sed -i -- 's/\${file(deployment-env.yml):artifact}/${config['artifact']}/g' serverless.yml"
@@ -538,6 +541,7 @@ def writeServerlessFile(config, environment_logical_id){
 	if (config['mainClass']) {
 		sh "sed -i -- 's/\${file(deployment-env.yml):mainClass}/${config['mainClass']}/g' serverless.yml"
 	}
+
 
 }
 
--- a/builds/jenkins-build-pack-api/Jenkinsfile
+++ b/builds/jenkins-build-pack-api/Jenkinsfile
@@ -531,9 +531,8 @@ def writeServerlessFile(config, environment_logical_id){
 	sh "sed -i -- 's/\${file(deployment-env.yml):providerTimeout}/${config['providerTimeout']}/g' serverless.yml"
 	sh "sed -i -- 's/\${file(deployment-env.yml):securityGroupIds}/${config['securityGroupIds']}/g' serverless.yml"
 	sh "sed -i -- 's/\${file(deployment-env.yml):subnetIds}/${config['subnetIds'] }/g' serverless.yml"
-  sh "sed -i -- 's|\${file(deployment-env.yml):eventScheduleRate}|${config['eventScheduleRate']}|g' serverless.yml"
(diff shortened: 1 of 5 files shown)


devsatishm:
Can you update the description for the PR? What changes are we proposing with this PR?

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -531,6 +531,9 @@ def writeServerlessFile(config, environment_logical_id){
 	sh "sed -i -- 's/\${file(deployment-env.yml):providerTimeout}/${config['providerTimeout']}/g' serverless.yml"
 	sh "sed -i -- 's/\${file(deployment-env.yml):securityGroupIds}/${config['securityGroupIds']}/g' serverless.yml"
 	sh "sed -i -- 's/\${file(deployment-env.yml):subnetIds}/${config['subnetIds'] }/g' serverless.yml"
+  sh "sed -i -- 's|\${file(deployment-env.yml):eventScheduleRate}|${config['eventScheduleRate']}|g' serverless.yml"
```
devsatishm: Why are these changes required in api build pack? Shouldn't we deal with these in lambda build pack instead?

builds/serverless-config-pack/serverless-java.yml
```diff
@@ -87,28 +87,30 @@ functions:
...
     #Start:isDynamoDbEnabled
     customEventRole: ${file(policyFile.yml):eventSubscribed}
     #End:isDynamoDbEnabled
-    #Start:isSqsEnabled
+    #Start:isSQSEventEnabled
     customEventRole: ${file(policyFile.yml):sqsRole}
-    #End:isSqsEnabled
-    #Start:isSqsEnabled
+    #End:isSQSEventEnabled
+    #Start:isSQSEventEnabled
```
devsatishm: Why is #Start:isSQSEventEnabled added again?

builds/jenkins-build-pack-api/Jenkinsfile
devsatishm: If these are absolutely required, I would suggest removing the section instead of replacing with empty values (like in the case above).

builds/serverless-config-pack/serverless-java.yml
SiniWilson: The comment section added here is to remove the section when it is not using. The comment previously added was   #Start:isSqsEnabled   #End:isSqsEnabled But in lambda , used #Start:isSQSEventEnabled   #End:isSQSEventEnabled  to remove.  In order to fix that, I had renamed the comment.

builds/jenkins-build-pack-api/Jenkinsfile
SiniWilson: Done. Removed the event scheduled as it is not required in API

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -531,14 +531,15 @@ def writeServerlessFile(config, environment_logical_id){
 	sh "sed -i -- 's/\${file(deployment-env.yml):providerTimeout}/${config['providerTimeout']}/g' serverless.yml"
 	sh "sed -i -- 's/\${file(deployment-env.yml):securityGroupIds}/${config['securityGroupIds']}/g' serverless.yml"
 	sh "sed -i -- 's/\${file(deployment-env.yml):subnetIds}/${config['subnetIds'] }/g' serverless.yml"
+  sh "sed -i -- '/#Start:resources/,/#End:resources/d' ./serverless.yml"
+  sh "sed -i -- '/#Start:isScheduleEnabled/,/#End:isScheduleEnabled/d' ./serverless.yml"
```
devsatishm: How about other sections like `isDynamoDbEnabled`, `isSQSEventEnabled` etc. (actually everything under `eventsDisabled`)? Those are not required in API build workflows. Why are we only removing `isScheduleEnabled` section?

builds/serverless-config-pack/serverless-java.yml
devsatishm: I'm looking at code fragment below - 
`#End:isSQSEventEnabled`   
`#Start:isSQSEventEnabled`

Will the below work - ?

    `#Start:isSQSEventEnabled`
    `customEventRole: ${file(policyFile.yml):sqsRole}`
    `#Start:isSqsResourceNotExist`
    `{event_source_sqs}:`
      `Type: AWS::SQS::Queue`
      `Properties:`
        `QueueName: {event_source_sqs}`
        `VisibilityTimeout: 165`
    `#End:isSqsResourceNotExist`
    `#End:isSQSEventEnabled`

SiniWilson: Oh ok understood your point. 
The SQS flow is , first  will add the sqs queue name in the resource part and will deploy, if the queue not exists, serverless will create the resource - by using the serverless yml file and will attach the lambda arn to the queue. Deployment will be success in this case.
2. If the queue already exists - Serverless will add the queue name in the resource part and will deploy, this time we will get an error while deploying saying resource already exists. So I will capture that error and remove the resource part from the serverless yml file and will redploy. So for removing the resource part only, I had added that comment section, since we need the custom role for both the case

PR #544, opened 2018-09-05
Added assets for apigee and fixed lambda asset issue

svidhani:
Fixed the issue with assets not getting created and updated the build pack module to call events properly. Validated using assets lambda at this point.

### DIFF
--- a/builds/jenkins-build-pack-api/Jenkinsfile
+++ b/builds/jenkins-build-pack-api/Jenkinsfile
@@ -217,6 +217,7 @@ node() {
 			if (env_metadata["AWS_API_ID"] != null) {
 				aws_api_id = env_metadata["AWS_API_ID"]
 			}
+      def swaggerDocUrl=""
 
 			try {
 
@@ -241,7 +242,7 @@ node() {
 					sh "serverless deploy --stage ${environment_logical_id} -v --bucket ${configLoader.AWS.S3[envBucketKey]}"
 
 					def lambdaARN = awsLambdaModule.getLambdaARN(stackName);
-					events.sendCompletedEvent('CREATE_ASSET', null, utilModule.generateAssetMap("aws", lambdaARN, "lambda", config), environment_logical_id);
+					events.sendCompletedEvent('CREATE_ASSET', null, utilModule.generateAssetMap("aws", lambdaARN.arn, "lambda", config), environment_logical_id);
 
 					if (fileExists('swagger/swagger.json')) {
 						echo "Generating swagger file for environment: ${env_key}"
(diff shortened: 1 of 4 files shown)


builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -291,24 +292,16 @@ node() {
...
 								events.sendStartedEvent('DEPLOY_TO_GCP_APIGEE', 'Deployment started for APIGEE', context_map, environment_logical_id)
-								def arn = awsLambdaModule.getLambdaARN(stackName)
 								apiHostName ="${configLoader.APIGEE.API_ENDPOINTS[env_key].service_hostname}"
 								updateSwagger(apiHostName, environment_logical_id, current_environment, env_key, config)
-								endpointUrl = apigeeModule.deploy("swagger/swagger.json", arn, env_key, environment_logical_id, config, context_map)
+								endpointUrl = apigeeModule.deploy("swagger/swagger.json", lambdaARN, env_key, environment_logical_id, config, context_map)
 								echo "API Proxy created at: ${endpointUrl}"
-                events.sendCompletedEvent('CREATE_ASSET', null, utilModule.generateAssetMap("gcp", null, "apigee", config), environment_logical_id);
 								events.sendCompletedEvent('DEPLOY_TO_GCP_APIGEE', 'Deployment complete for APIGEE', context_map, environment_logical_id)
+                events.sendCompletedEvent('CREATE_ASSET', null, utilModule.generateAssetMap("gcp", endpointUrl, "apigee", config), environment_logical_id);
```
devsatishm: Asset type should be endpoint_url  instead of apigee (since you are passing endpointUrl as value).

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -291,24 +292,16 @@ node() {
...
 								break;
 							default:
 								error "Deployment targets are not defined"
 
-              def apiArns = getAPIGatewayResourceARNs(aws_api_id, environment_logical_id)
```
devsatishm: Why did we remove this? Should we move to aws section instead?

svidhani: Its already there in aws api gateway(expand the view). Seems like this was not removed.

core/jazz_assets/config/global-config.json
```diff
@@ -30,7 +30,8 @@
 		"dynamodb",
 		"cloudwatch_event",
 		"swagger_url",
-		"endpoint_url"
+		"endpoint_url",
+    "apigee"
```
devsatishm: apigee can't be an asset type.

builds/jenkins-build-pack-api/Jenkinsfile
svidhani: i am passing enpoint_url as it needs some value and i can't pass arn. Passing null to provider_id will throw an error and not create the asset.

core/jazz_assets/config/global-config.json
svidhani: updated

builds/jenkins-build-pack-api/Jenkinsfile
svidhani: updated

PR #565, opened 2018-09-29
Followup PR for splunk

supritat:
### Requirements

* Bug fixes for splunk and kinesis log streamer
* Including unit test for jazz_es-kinesis-log-streamer

### Description of the Change
Small fixes are done after QA.


core/jazz_es-kinesis-log-streamer/index.js
```diff
@@ -86,25 +91,24 @@ function handler(input, context, cb) {
       return cb(null, responseObj("Success", input));
     }
   } catch (err) {
-    logger.error(err);
-    return cb(JSON.stringify(errorHandler.throwInternalServerError(JSON.stringify(err.message))));
+    logger.error("Something went wrong. " + err);
```
svidhani: instead of something went wrong. Say something like "Transform error" or something and then err.

PR #594, opened 2018-10-25
bug fix for splunk enable condition

supritat:
### Requirements

### DIFF
--- a/core/jazz_splunk-kinesis-log-streamer/index.js
+++ b/core/jazz_splunk-kinesis-log-streamer/index.js
@@ -36,7 +36,7 @@ function handler(event, context, callback) {
   logger.debug('Received event:' + JSON.stringify(event));
   let errorHandler = errorHandlerModule();
   let config = configData.getConfig(event, context);
-  let isSplunkEnabled = (global_config.ENABLE_SPLUNK).toLowerCase() == 'true' ? true : false;
+  let isSplunkEnabled = (global_config.ENABLE_SPLUNK && (global_config.ENABLE_SPLUNK).toLowerCase()) == 'true' ? true : false;
 
   if (isSplunkEnabled && event.Records && event.Records.length) {
     const loggerConfig = {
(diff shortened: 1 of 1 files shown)


core/jazz_splunk-kinesis-log-streamer/index.js
```diff
@@ -36,8 +36,9 @@ function handler(event, context, callback) {
   logger.debug('Received event:' + JSON.stringify(event));
   let errorHandler = errorHandlerModule();
   let config = configData.getConfig(event, context);
+  let isSplunkEnabled = (global_config.ENABLE_SPLUNK).toLowerCase() == 'true' ? true : false;
```
devsatishm: Check if global_config.ENABLE_SPLUNK exists & is a string before converting it into lowercase. Return 5XX (since it's a service configuration error) if the value is invalid/empty.

supritat: Updated.

core/jazz_splunk-kinesis-log-streamer/index.js
```diff
@@ -36,8 +36,9 @@ function handler(event, context, callback) {
   logger.debug('Received event:' + JSON.stringify(event));
   let errorHandler = errorHandlerModule();
   let config = configData.getConfig(event, context);
+  let isSplunkEnabled = (global_config.ENABLE_SPLUNK && (global_config.ENABLE_SPLUNK).toLowerCase() == 'true') ? true : false;
```
devsatishm: To handle boolean & string cases, I would recommend this - 
```suggestion
  let isSplunkEnabled = (global_config.ENABLE_SPLUNK && (global_config.ENABLE_SPLUNK).toString().toLowerCase() === 'true') ? true : false;
```

core/jazz_splunk-kinesis-log-streamer/index.js
```diff
@@ -36,8 +36,9 @@ function handler(event, context, callback) {
   logger.debug('Received event:' + JSON.stringify(event));
```
devsatishm: For some reason, you are missing `logger.init` before using `logger`
```suggestion
  logger.init (event, context));
  logger.debug('Received event: ' + JSON.stringify(event));
```

core/jazz_splunk-kinesis-log-streamer/index.js
supritat: updated

core/jazz_splunk-kinesis-log-streamer/index.js
supritat: Included

PR #598, opened 2018-10-30
Admin Unit Tests

pspani:
Requirements
adding unit tests to admin api

Description of the Change
added unit tests to the jazz_admin service module with 90% coverage

Benefits
the module was missing unit tests.

Possible Drawbacks
Some minor changes were made to the jazz_admin module for enabling the unit tests. This would need to be tested after deploying.

Applicable Issues

### DIFF
--- a/core/jazz_admin/test/test.js
+++ b/core/jazz_admin/test/test.js
@@ -25,7 +25,6 @@ const configObj = require('../components/config.js');
 describe('jazz_admin', function () {
 
   beforeEach(function () {
-    spy = sinon.spy();
     event = {
       "stage": "test",
       "method": "GET",
@@ -78,7 +77,7 @@ describe('jazz_admin', function () {
 
     it("should indicate error while making request to gitlab repo", () => {
       config.SCM_TYPE = "gitlab";
-      var responseObj = {
+      let responseObj = {
         statusCode: 401,
         body: {
           data: {},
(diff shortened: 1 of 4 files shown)


core/jazz_admin/test/test.js
```diff
@@ -1,11 +1,171 @@
...
+const request = require('request');
+
 const index = require('../index');
+const awsContext = require('aws-lambda-mock-context');
+const configObj = require('../components/config.js');
+
+describe('jazz_admin', function () {
+
+  beforeEach(function () {
+    spy = sinon.spy();
```
svidhani: why do you need this? r u using it anywhere?

core/jazz_admin/test/test.js
```diff
@@ -1,11 +1,171 @@
...
+        expect(error).to.include('{"errorType":"Unauthorized","message":"This user is not authorized to access this service.');
+      });
+    });
+  });
+
+  describe('getInstallerVarsJSON', () => {
+
+    it("should indicate error while making request to gitlab repo", () => {
+      config.SCM_TYPE = "gitlab";
+      var responseObj = {
```
svidhani: try to use const/let, etc.

core/jazz_admin/test/test.js
```diff
@@ -1,11 +1,171 @@
...
+        }
+      };
+      reqStub = sinon.stub(request, "Request").callsFake((obj) => {
+        return obj.callback(null, responseObj, responseObj.body)
+      });
+      index.getInstallerVarsJSON(config)
+        .catch(error => {
+          expect(error).to.include('Unauthorized');
+        });
+      sinon.assert.calledOnce(reqStub);
```
svidhani: you should put these statements(sinon.assert and restore inside the callback and not outside.

`.catch(error => {
          expect(error).to.include('Unauthorized');
          sinon.assert.calledOnce(reqStub);
          reqStub.restore();
  })`


fix it everywhere

core/jazz_admin/test/test.js
pspani: removed. I had initially thought I needed it but forgot to remove it when I didn't need it.

PR #600, opened 2018-10-30
Fix vulnerability in the Python request pacakge

suryajak:
### Description of the Change
Fix vulnerability with the current `request` package that we are using in our templates: https://nvd.nist.gov/vuln/detail/CVE-2018-18074

### Benefits

Users creating python based services will be using secure package versions

### Possible Drawbacks
None


suryajak:
Validated that the template works with the newer version.

PR #601, opened 2018-10-31
Resource per environment support for event source dynamodb

SiniWilson:
### Requirements

Resource per environment support for event source dynamodb

### Description of the Change

Created dynamodb table as part of dynamodb event subscribed services
Created  new stream if not exists as part of dynamodb event subscribed services
Asset table get updated for the newly created resources

### Benefits

User can create new dynamodb table as part of dynamodb event subscribed services
User can view assets  as part of dynamodb event subscribed services

### DIFF
--- a/builds/serverless-config-pack/serverless-java.yml
+++ b/builds/serverless-config-pack/serverless-java.yml
@@ -49,7 +49,20 @@ functions:
           enabled: ${file(deployment-env.yml):eventScheduleEnable}
       #End:isScheduleEnabled
       #Start:isDynamoDbEnabled
-      - stream: {event_source_dynamodb}
+      - stream:
+          type: dynamodb
+          #Start:dynamoDbstreamArn
+          streamArnDisabled: {event_dynamodb_stream_arn}
+          #End:dynamoDbstreamArn
+          #Start:dynamoDbstreamGetArn
+          arn:
+            Fn::GetAtt:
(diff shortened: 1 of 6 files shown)


builds/serverless-config-pack/serverless-java.yml
```diff
@@ -112,5 +125,23 @@ resourcesDisabled:
...
         ShardCount: 1
     #End:isStreamEnabled
+    #Start:isDynamoDbEnabled
+    #Start:isDynamoDbtableNotExist
+    sourceDynamoDbTable:
+       Type: AWS::DynamoDB::Table
+       Properties:
+         TableName: {event_source_dynamodb}
+         AttributeDefinitions:
+           - AttributeName: ID
```
svidhani: This seems to be a change in the requirement. Earlier the story was only to enable stream(if not already done). Now  I am seeing we are getting into creating the table(if it does not exist). The issue here is that partition key has to be given when creating it and it cannot be changed. You can add or delete secondary indexes. @devsatishm when did this requirement change and are we ok with this

devsatishm: Good point. We are bringing in the ability to create tables with few (_hardcoded_) configurations to begin with. This will change later and developer is expected to provide relevant information (like participation key for example) during service creation through UI/CLI or using some kind of application template configurations.

PR #608, opened 2018-11-09
Delete contents inside S3 bucket before deleting it.

ajay1kumar1:
### Requirements

Jazz gives error while deleting services of type: function with S3 configured as event trigger.

### Description of the Change

Empty S3 bucket (delete the content inside it) before deleting the assets created by Jazz (function, S3 bucket etc) when S3 is configured as event trigger for the service.

### DIFF
--- a/builds/delete-serverless-service-build-pack/Jenkinsfile
+++ b/builds/delete-serverless-service-build-pack/Jenkinsfile
@@ -278,8 +278,14 @@ def checkBucket(stage) {
 			if(stage != "prod"){
 				targetBucket = service_config['event_source_s3'] + "-" + stage;
 			}
-			def s3Empty = isBucketEmpty(targetBucket);											
-			if(!s3Empty) {
+      def s3Exists = true;
+        try {
+            sh "aws s3api head-bucket --bucket ${targetBucket} --output json"
+        } catch (ex) {
+            echo "Bucket does not exist"
+            s3Exists = false
+        }
+      if(s3Exists) {
 				def output = sh(
 					script: "aws s3 ls s3://${targetBucket} --recursive | wc -l",
 					returnStdout: true
--- a/builds/delete-serverless-service-build-pack/Jenkinsfile
+++ b/builds/delete-serverless-service-build-pack/Jenkinsfile
@@ -278,8 +278,8 @@ def checkBucket(stage) {
 			if(stage != "prod"){
 				targetBucket = service_config['event_source_s3'] + "-" + stage;
 			}
-			def s3Exists = isBucketEmpty(targetBucket);											
-			if(s3Exists) {
+			def s3Empty = isBucketEmpty(targetBucket);											
+			if(!s3Empty) {
 				def output = sh(
 					script: "aws s3 ls s3://${targetBucket} --recursive | wc -l",
 					returnStdout: true
--- a/builds/delete-serverless-service-build-pack/Jenkinsfile
+++ b/builds/delete-serverless-service-build-pack/Jenkinsfile
@@ -279,12 +279,12 @@ def checkBucket(stage) {
 				targetBucket = service_config['event_source_s3'] + "-" + stage;
 			}
       def s3Exists = true;
-        try {
-            sh "aws s3api head-bucket --bucket ${targetBucket} --output json"
-        } catch (ex) {
-            echo "Bucket does not exist"
-            s3Exists = false
-        }
+      try {
+          sh "aws s3api head-bucket --bucket ${targetBucket} --output json"
+      } catch (ex) {
+          echo "Bucket does not exist"
+          s3Exists = false
+      }
      if(s3Exists && (!isBucketEmpty(targetBucket))) {				
 					sh "aws s3 rm s3://${targetBucket}/ --recursive --exclude '/'"
 					echo "Removing items from bucket for ${output}"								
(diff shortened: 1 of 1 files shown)


builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -194,6 +194,10 @@ node {
 							}
 							events.setEnvironment(_envId)
 							events.sendStartedEvent('DELETE_ENVIRONMENT', "Environment cleanup for " + _envId + " started", null, _envId)
+              echo 'Service deletion is requested: ' + current_environment_id
+							if (is_service_deletion && service_config['event_source_s3']) { // check service for S3 Bucket and make empty if not. //
```
SiniWilson: what will be happen if an environment get deleted and the bucket is not empty?

builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -255,6 +259,47 @@ node {
...
+				sh "aws s3api head-bucket --bucket ${targetBucket} --output json"
+			} catch (ex) {
+				echo "Bucket does not exist"
+				s3Exists = false
+			}								
+			if(s3Exists) {
+				def output = sh(
+					script: "aws s3 ls s3://${targetBucket} --recursive | wc -l",
+					returnStdout: true
+				).trim()
```
SiniWilson: There are existing functions for checking bucket exist or not and for the bucket is empty or not. Kindly reuse the existing methods.

builds/delete-serverless-service-build-pack/Jenkinsfile
ajay1kumar1: That is the actual issue, SLS unable to delete environment, when bucket is not empty.
i.e environment will not delete until bucket is empty in S3 case.

SiniWilson: Actually, my concern is,  We are manually deleting the content of the bucket, **only if (is_service_deletion && service_config['event_source_s3']) {** this condition is satisfying. Did you checked, this feature by deleting an environment ?

builds/delete-serverless-service-build-pack/Jenkinsfile
ajay1kumar1: updated

builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -255,6 +259,42 @@ node {
...
+			sh "aws configure set profile.cloud-api.region ${configLoader.AWS.REGION}"
+			sh "aws configure set profile.cloud-api.aws_access_key_id $AWS_ACCESS_KEY_ID"
+			sh "aws configure set profile.cloud-api.aws_secret_access_key $AWS_SECRET_ACCESS_KEY"					
+			
+			// Empty bucket
+			def targetBucket = service_config['event_source_s3'];
+			if(stage != "prod"){
+				targetBucket = service_config['event_source_s3'] + "-" + stage;
+			}
+			def s3Exists = isBucketEmpty(targetBucket);
```
SiniWilson: isBucketEmpty as  the name says, it will check whether the bucket empty or not. First check bucket exists or not and then bucket empty. We can take this offline, if you have need any clarification

builds/delete-serverless-service-build-pack/Jenkinsfile
SiniWilson: Let us take this offline

builds/delete-serverless-service-build-pack/Jenkinsfile
svidhani: this logic is wrong. is BucketEmpty is giving you the value of false if there are contents in the bucket. You are then using that assuming it will be true to run the logic to delete. But what you are getting is value of false. You need to change the variable name to say something like s3Empty and then check for if (!s3Empty)

builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -255,6 +259,48 @@ node {
...
+				targetBucket = service_config['event_source_s3'] + "-" + stage;
+			}
+      def s3Exists = true;
+        try {
+            sh "aws s3api head-bucket --bucket ${targetBucket} --output json"
+        } catch (ex) {
+            echo "Bucket does not exist"
+            s3Exists = false
+        }
+     if(s3Exists && (!isBucketEmpty(targetBucket))) {
```
svidhani: here is the code you need

try {
                sh "aws s3api head-bucket --bucket ${targetBucket} --output json"
            } catch (ex) {
                echo "Bucket does not exist"
                s3Exists = false
            }
			if(s3Exists) {
				def s3Empty = isBucketEmpty(targetBucket)
				if(!s3Empty){
					sh "aws s3 rm s3://${targetBucket}/ --recursive --exclude '/'"
					echo "Removing items from bucket for ${output}"
				}					
			}			
		} catch (ex) {
			handleFailureEvent(ex.getMessage())
		}

builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -255,6 +259,42 @@ node {
...
+			sh "aws configure set profile.cloud-api.aws_access_key_id $AWS_ACCESS_KEY_ID"
+			sh "aws configure set profile.cloud-api.aws_secret_access_key $AWS_SECRET_ACCESS_KEY"					
+			
+			// Empty bucket
+			def targetBucket = service_config['event_source_s3'];
+			if(stage != "prod"){
+				targetBucket = service_config['event_source_s3'] + "-" + stage;
+			}
+      def s3Exists = true;
+        try {
```
svidhani: indentation?

PR #616, opened 2018-11-13
Handling asset updation and resources deletion

SiniWilson:
### Requirements

Handling asset updation and the deletion of the resources , that jazz created by using aws cli

### Description of the Change

Updated the asset catalog
Deleted the resources that jazz created by using aws cli

### Benefits

The resources get deleted/disabled as part of the service deletion


svidhani:
Please resolve conflicts

PR #619, opened 2018-11-14
:bug: Deleting event notification configuration as part of the service deletion

SiniWilson:
### Requirements

Deleting s3 event notification configuration as part of the service deletion

### Description of the Change

Get all the existing event notification configuration for the s3 bucket
Update the configuration for the function trigger

### Benefits

The event notification trigger for the service get deleted as part of the service deletion

### DIFF
--- a/builds/jazz-build-module/aws-lambda-events-module.groovy
+++ b/builds/jazz-build-module/aws-lambda-events-module.groovy
@@ -137,13 +137,13 @@ def removeS3EventsFromServerless(isEventSchdld){
 
 def checkS3BucketExists(s3BucketName){
   try {
-    sh "aws s3api head-bucket --bucket $s3BucketName --output json"
+    sh "aws s3api head-bucket --bucket $s3BucketName --profile cloud-api --output json"
     echo "Bucket exists and have access"
     return true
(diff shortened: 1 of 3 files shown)


builds/jazz-build-module/aws-lambda-events-module.groovy
```diff
@@ -137,13 +137,13 @@ def removeS3EventsFromServerless(isEventSchdld){
...
   try {
-    sh "aws s3api head-bucket --bucket $s3BucketName --output json"
+    sh "aws s3api head-bucket --bucket $s3BucketName --profile cloud-api --output json"
     echo "Bucket exists and have access"
     return true
   } catch (ex) {//bucket exists but with no access
     def res
     try {
-      res = sh(script: "aws s3api head-bucket --bucket $s3BucketName --output json 2<&1 | grep -c 'Forbidden'", returnStdout: true).trim()
+      res = sh(script: "aws s3api head-bucket --bucket $s3BucketName --profile cloud-api --output json 2<&1 | grep -c 'Forbidden'", returnStdout: true).trim()
```
svidhani: u can skip the variable and return directly from here. no need for if condition below

PR #621, opened 2018-11-15
Random aws profile generation 

glearner:
### Requirements

* cloud-api used as aws profile .. so i changed it to random aws profile.

### Description of the Change

Now it uses cloud-api as aws profile names during builds.  There is a possibility of two builds running at the same time using the same aws profile  on the same agent. One of the builds can clear out the profile while the other build is in progress and this will lead to Client Token expired error. 

### Benefits

 Client Token expired error resolved.

### Possible Drawbacks

NONE

### Applicable Issues

NA


svidhani:
please change the base to develop and resolve the conflicts

glearner:
done

PR #635, opened 2018-11-26
Refresh changes

mayajohns:
### Requirements

Verification of refresh button fix-Browser refresh happens affected the UI refresh

CSS path also changed to xpath for 2 since its not getting identified by browser on load time

### DIFF
--- a/core/jazz_ui/e2e/04-jazzservices/jazzservices.e2e-spec.ts
+++ b/core/jazz_ui/e2e/04-jazzservices/jazzservices.e2e-spec.ts
@@ -66,8 +66,7 @@ describe('Overview', () => {
             jazzServices_po.getAwsServiceName().click();
             browser.wait(EC.visibilityOf(jazzServices_po.getServiceNameHeader()), timeOutHigh);
             browser.sleep(45000);
-            browser.refresh();
-            browser.driver.switchTo().activeElement();
+            browser.wait(EC.elementToBeClickable(jazzServices_po.getRefresh()), timeOutHigh);
             jazzServices_po.getRefresh().click();
             browser.sleep(30000);
       });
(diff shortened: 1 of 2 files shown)


core/jazz_ui/e2e/page-objects/jazzservices.po.ts
```diff
@@ -108,5 +109,11 @@ export class Jazz {
   getAssetStatusVerify(){
     return element(by.xpath('//li[@class="x caps active"]'));
   }
+  getLamdaName(){
+    return element(by.xpath('//div[@class="table-row pointer"]/div[text()="function"]/preceding-sibling::div'));
```
svidhani: is this guaranteed?
this is more like a hacky way to capturing the name

core/jazz_ui/e2e/04-jazzservices/jazzservices.e2e-spec.ts
```diff
@@ -118,6 +117,49 @@ describe('Overview', () => {
...
+      it('Verify Lamda Function Page Title', () => {
+            browser.wait(EC.visibilityOf(jazzServices_po.getAwsServiceName()), timeOutHigh);
+            browser.wait(EC.elementToBeClickable(jazzServices_po.getAwsServiceName()), timeOutHigh);
+            jazzServices_po.getLamdaName().click();
+            expect(jazzServices_po.getOverviewStatus().getText()).toEqual('OVERVIEW');
+            browser.wait(EC.visibilityOf(jazzServices_po.getServiceNameHeader()), timeOutHigh);
+            browser.sleep(45000);
+            browser.wait(EC.elementToBeClickable(jazzServices_po.getRefresh()), timeOutHigh);
+            jazzServices_po.getRefresh().click();
+            browser.sleep(30000);
```
svidhani: don't you think you can add something like afterEach where you can put a timer instead along with other cleanup stuff (that needs to be run after every test)?

core/jazz_ui/e2e/page-objects/jazzservices.po.ts
mayajohns: yes Sid Its guaranteed since we are taking the Lamda name with the text "function" under the corresponding table names

core/jazz_ui/e2e/04-jazzservices/jazzservices.e2e-spec.ts
mayajohns: what the purpose of Refresh here is we have one button in UI that needs to be clicked inorder to get the Prod and Stg

core/jazz_ui/e2e/04-jazzservices/jazzservices.e2e-spec.ts
```diff
@@ -118,6 +117,49 @@ describe('Overview', () => {
...
+      });
+
+      it('Verify Lamda Prod Name' ,  () => {
+            browser.wait(EC.visibilityOf(jazzServices_po.getServiceNameHeader()), timeOutHigh);
+            jazzServices_po.getProdName().click();
+            expect(jazzServices_po.getDeploymentStatus().getText()).toEqual('DEPLOYMENTS');
+            browser.wait(EC.visibilityOf(jazzServices_po.getProdHeader()), timeOutHigh);
+            browser.wait(EC.visibilityOf(jazzServices_po.getRefresh()), timeOutHigh);
+            browser.driver.switchTo().activeElement();
+            browser.sleep(15000);
```
pspani: these browser.sleep values work at the moment on your PC with network delays that make sense from your place to work. they may need to be tweaked if run from some other server. Putting these values in the config file would help in tweaking these values at runtime. Do you agree?

core/jazz_ui/e2e/page-objects/jazzservices.po.ts
```diff
@@ -76,7 +76,8 @@ export class Jazz {
      return element(by.css('section.footer-btn>btn-jazz-primary'));
     }
   getServiceNameHeader() {
-     return element(by.css('div.page-title-wrap.hide-for-mobile>h1.page-hdr.bold-title'));
+     
+	 return element(by.xpath('//div[@class="page-title-wrap hide-for-mobile"]/h1[@class="page-hdr bold-title"]'));
```
pspani: I wonder why the browser was not finding these elements on load. Were you able to find more details on this? Just curious. 
But I agree with you that for now, we can go with xpath where finding by element id is not trivial.

core/jazz_ui/e2e/04-jazzservices/jazzservices.e2e-spec.ts
mayajohns: Parth,why we are giving sleep is like,In our application For eg:one scenario,after clicking the API,UI takes some time to get it build and validate .after that only we will get the Prod properly.Its not kept to adjust with my System performance.Each and every place we are giving time for the Reg verification mail,login and deployment etc.I have a discussion with Sid about the changes of this sleep,more standard way we can give explicit values,but that we can go for a separate story with the miscellaneous items.

core/jazz_ui/e2e/page-objects/jazzservices.po.ts
mayajohns: I was trying both Parth,but may be some browser specific issues,I didn't go in deep since I am already running out of time

pspani: understood... no problems

core/jazz_ui/e2e/04-jazzservices/jazzservices.e2e-spec.ts
pspani: understood.. lets take it up as part of a separate story

PR #671, opened 2018-12-13
Replace slider with dropdown in advanced-filters component

supritat:
### Requirements


suryajak:
> > Isn't this a fix for the UI issue from splunk branch? If yes, why is this going to develop? The PR should be for splunk or master branch.
> 
> `develop` is now in sync with `master` (which means splunk changes got into `develop` already). It's broken now and #672 should fix it.
Fix is required for `master`. It's good `develop` is in sync with `master` but it has other features in it which needs to be tested and validated (perhaps one of the next releases).

PR #681, opened 2018-12-18
ACL policies implementation

svidhani:
### Requirements

Implemented the ACL policies
- GET
- POST -> add and delete

### DIFF
--- a/core/jazz_acl/components/casbin.js
+++ b/core/jazz_acl/components/casbin.js
@@ -17,55 +17,165 @@
 const Enforcer = require('casbin').Enforcer;
 const TypeORMAdapter = require('typeorm-adapter');
 const logger = require('./logger.js');
+const errorHandlerModule = require("./error-handler.js")();
 
+/* Create a connection to the DB*/
 async function dbConnection(config) {
-  const conn = await TypeORMAdapter.default.newAdapter({
-    type: config.CASBIN.TYPE,
-    host: config.CASBIN.HOST,
-    port: config.CASBIN.PORT,
-    username: config.CASBIN.USER,
-    password: config.CASBIN.PASSWORD,
-    database: config.CASBIN.DATABASE,
-    timeout: config.CASBIN.TIMEOUT
-  });
+  let conn;
+  try {
+    conn = await TypeORMAdapter.default.newAdapter({
+      type: config.CASBIN.TYPE,
+      host: config.CASBIN.HOST,
+      port: config.CASBIN.PORT,
+      username: config.CASBIN.USER,
+      password: config.CASBIN.PASSWORD,
+      database: config.CASBIN.DATABASE,
+      timeout: config.CASBIN.TIMEOUT
+    });
+  } catch(err) {
+    if (err.name !== "AlreadyHasActiveConnectionError") {
+      logger.error(err.message);
+      throw (errorHandlerModule.throwInternalServerError(err.message));
+    }
+  }
 
-  const enforcer = await Enforcer.newEnforcer('../config/rbac_model.conf', conn);
-  return enforcer;
+  return conn;
 }
 
-async function addPermissionForUser(serviceId, policies) {
-  //TODO implement addPermissionForUser
+/* Get the policies from casbin given the serviceId*/
+async function getPolicies(serviceId, config) {
+  const values = [`${serviceId}_manage`, `${serviceId}_code`, `${serviceId}_deploy`];
+  const result = await getFilteredPolicy(1, values, config);
+  if (result && result.success) {
+    let filteredPolicies = result.filter(el => el.length >=1);
+    if (filteredPolicies.length) {
+      filteredPolicies = filteredPolicies.map(policyArr => policyArr.map(policy => [policy[0], policy[1].split('_')[1], policy[2]]));
+      result = filteredPolicies;
+    } else {
+      result = [];
+    }
+  }
 
-  return true;
-}
-
-async function enforce(policy) {
-
-  //TODO implement the enforce method
-  return true;
-}
-
-/* Get the policies from casbin given the value*/
-async function getFilteredPolicy(index, values, config) {
-  //TODO Implement getFilteredPolicy
   return result;
 }
 
-async function removeFilteredPolicy(index, value) {
-  //TODO: implement removeFilteredPolicy
+/* Get the policies from casbin given the values and index*/
+async function getFilteredPolicy(index, values, config) {
+  let result = {};
+  let conn, enforcer;
 
-  return true;
+  try {
+    conn = await dbConnection(config);
+    enforcer = await Enforcer.newEnforcer("./config/rbac_model.conf", conn);
+
+    const promisedPolicies = values.map(async value => await enforcer.getFilteredPolicy(index, value));
+    const policies = await Promise.all(promisedPolicies);
+    result = policies;
+  } catch(err) {
+    logger.error(err.message);
+    result.error = err.message;
+  } finally {
+    await conn.close();
+  }
+
+  return result;
 }
 
-async function getPermissionForUser(user) {
-  //TODO: implement getPermissionForUser
+/* Check permissions for a user */
+async function checkPermissions(userId, serviceId, category, permission, config) {
+  let result = {};
 
-  return [];
+  try {
+    const conn = await dbConnection(config);
+    const enforcer = await Enforcer.newEnforcer('./config/rbac_model.conf', conn);
+    result.authorized = enforcer.enforce(userId, `${serviceId}_${category}`, permission);
+  } catch(err) {
+    logger.error(err.message);
+    result = {
+      authorized: false,
+      error: err.message
+    };
+  } finally {
+    await conn.close();
+  }
+
+  return result;
+}
+
+/* Add and/or Remove filtered policy */
+async function addOrRemovePolicy(serviceId, config, action, policies) {
+  let result = {};
+  let conn;
+  const objects = [`${serviceId}_manage`, `${serviceId}_code`, `${serviceId}_deploy`];
+  let totalPolices = 0;
+  try {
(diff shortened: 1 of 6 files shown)


core/jazz_acl/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
```
devsatishm: Can you update  other default values that would have come with the template - name, desc, main etc.?

core/jazz_acl/components/error-handler.js
```diff
@@ -18,30 +18,35 @@ module.exports = () => {
     const errorObj = {
         throwInputValidationError: (errorMessage) => (
             {
+                errorCode: 400,
```
devsatishm: I  think `errorCode` here is not necessary here. API Gateway does a mapping of `errorType` value to HTTP status code.

core/jazz_acl/index.js
```diff
@@ -24,55 +23,94 @@ const casbinUtil = require("./components/casbin.js");
...
-    await validation.validateInput(event);
-    const aclResult = await processACL(event);
-
-    return responseObj(aclResult, event.body);
+    validation.validateBasicInput(event);
+    const aclResult = await processACLRequest(event, config);
 
+    return {
+      status: 200,
+      body: aclResult
```
devsatishm: - `status` is not required. 
-  key `body` doesn't exist in swagger. Are we sending payload as per the swagger spec?

core/jazz_acl/package.json
svidhani: updated

svidhani: updated

core/jazz_acl/components/error-handler.js
svidhani: updated

core/jazz_acl/index.js
svidhani: updated

core/jazz_acl/components/validation.js
```diff
@@ -16,23 +16,47 @@
...
-async function validateInput(userInput) {
+function validateBasicInput(userInput) {
 
 		if (!userInput || !userInput.method) {
 			throw (errorHandlerModule.throwInputValidationError("Invalid or missing arguments"));
     }
 
-    if (!userInput.resourcePath) {
+    if (!userInput.path) {
 			throw (errorHandlerModule.throwInputValidationError("Missing the resource path"));
```
devsatishm: `path` is usually set by API Gateway. If it's missing, it's more of a system integration error (`InternalServerError`)

core/jazz_acl/components/validation.js
```diff
@@ -16,23 +16,47 @@
 
 const errorHandlerModule = require("./error-handler.js")();
 
-async function validateInput(userInput) {
+function validateBasicInput(userInput) {
 
 		if (!userInput || !userInput.method) {
 			throw (errorHandlerModule.throwInputValidationError("Invalid or missing arguments"));
```
devsatishm: Please separate these out. 

- method is usually set by API Gateway. If it's missing, it's more of a system integration error (InternalServerError).
- if `userInput` is missing, it's `InputValidationError`

core/jazz_acl/components/validation.js
```diff
@@ -16,23 +16,47 @@
...
+    throw (errorHandlerModule.throwInputValidationError("Query parameters are missing"));
+  }
+
+  if (!userInput.query.serviceId) {
+    throw (errorHandlerModule.throwInputValidationError("Service Id is missing"));
+  }
+}
+
+function validatePostPoliciesInput(userInput) {
+  if (!userInput.body) {
```
devsatishm: `body` is usually set by API Gateway. If it's missing, it's more of a system integration error (`InternalServerError`)

core/jazz_acl/components/validation.js
svidhani: This logic will take care of empty/missing values as well.
i will update the one above(for method to remove userInput.

core/jazz_acl/components/casbin.js
```diff
@@ -17,55 +17,185 @@
...
+  }
 
-  return true;
+  return result;
 }
 
-async function getPermissionForUser(user) {
-  //TODO: implement getPermissionForUser
+/* Remove filtered policy */
+async function removeFilteredPolicy(serviceId, config) {
```
devsatishm: Suggestion - We're creating two DB connections during the execution of this function - `removeFilteredPolicy`. Is there a possibility to avoid this & reuse connections?

core/jazz_acl/components/casbin.js
```diff
@@ -17,55 +17,185 @@
...
+      };  
+    }
+  } catch(err) {
+    logger.error(err.message);
+    result = {
+      "success": false,
+      "error": err.message
+    };
+  }
+  await conn.close();
```
devsatishm: Can we move things like these to `finally` block (best practice?)

core/jazz_acl/index.js
```diff
@@ -14,8 +14,7 @@
 // limitations under the License.
 // =========================================================================
 
-const errorHandlerModule = require("./components/error-handler.js");
-const responseObj = require("./components/response.js");
+const errorHandlerModule = require("./components/error-handler.js")();
 const configModule = require("./components/config.js");
 const logger = require("./components/logger.js");
```
devsatishm: Logger should be updated as per https://github.com/tmobile/jazz/pull/695.

core/jazz_acl/components/casbin.js
```diff
@@ -17,55 +17,185 @@
...
+        removeResult = await Promise.all(removeResult);
+        await enforcer.savePolicy();
+        await conn.close();
+      }
+      
+      if (removeResult.length === totalPolicies) {
+        result.success = true;
+      } else {
+        result = {
+          "success": false,
```
devsatishm: `"success": false` is redundant (I think it is understood, since `error` already exists in the response).

core/jazz_acl/components/casbin.js
```diff
@@ -17,55 +17,185 @@
...
+    }
+  }
 
-  return true;
+  return result;
 }
 
-async function getPermissionForUser(user) {
-  //TODO: implement getPermissionForUser
+/* Remove filtered policy */
```
devsatishm: I see that this function is removing all polices for a service, while with `removeFilteredPolicy it sounds like we're removing specific polices. Should we rename this function?

core/jazz_acl/components/casbin.js
```diff
@@ -17,55 +17,185 @@
...
+async function getPermissionForUser(userId, config) {
+  const result = await getFilteredPolicy(0, userId, config);
+
+  return result;
 }
 
 module.exports = {
   addPermissionForUser,
-  getFilteredPolicy,
+  getPolicies,
```
devsatishm: `addPermissionForUser` is added twice.

core/jazz_acl/components/casbin.js
```diff
@@ -17,55 +17,185 @@
...
-  return enforcer;
+  return conn;
 }
 
-async function addPermissionForUser(serviceId, policies) {
-  //TODO implement addPermissionForUser
-
-  return true;
+/* Add and save the policy in casbin DB*/
+async function addPermissionForUser(serviceId, policies, config) {
```
devsatishm: We're making opening/closing DB connections thrice (two times in `removeFilteredPolicy` & one more directly). We should definitely avoid it.

core/jazz_acl/components/validation.js
```diff
@@ -16,23 +16,45 @@
...
+    throw (errorHandlerModule.throwInternalServerError("Body is missing"));
+  }
+
+  if (!userInput.body.serviceId) {
+    throw (errorHandlerModule.throwInputValidationError("Service Id is missing"));
+  }
 
-		return true;
+  if (!userInput.body.policies) {
```
devsatishm: Should we extend this and validate each policy object?  I see that we are using policy.userId, policy.category & policy.permission directly else where. It's better if we check for these k/v pairs if policies[] is not empty.

core/jazz_acl/components/casbin.js
devsatishm: We aren't adding permissions for a specific user here. How about `updateServicePermissions`?

core/jazz_acl/components/validation.js
svidhani: This is difficult to validate as policies can be empty for the same POST (add or delete)

core/jazz_acl/components/casbin.js
svidhani: i tried it and i have been getting issues with db connection. it is not closing it properly and even after lambda is executed the connection remains open. I even tried passing opened connection between function which was causing issues as well.

core/jazz_acl/components/validation.js
```diff
@@ -16,23 +16,63 @@
...
+      }
+      if (!policy.permission) {
+        missingFields.push("permission");
+      }
+      if (!policy.category) {
+        missingFields.push("category");
+      }
+    });
+    if (missingFields.length) {
+      throw (errorHandlerModule.throwInputValidationError(`Policy details are missing values for ${missingFields.join(', ')}`));
```
devsatishm: Since you are doing this for an array of policies, this message can get messy making it difficult the understand the real error (Ex: `Policy details are missing values for userId, userId, userId, permission, category, category`)

Let's keep it simple by saying these fields are required in each policy.

core/jazz_acl/package.json
```diff
@@ -10,10 +10,13 @@
...
     "casbin": "^1.1.6",
+    "chai-as-promised": "^7.1.1",
```
SiniWilson: "chai-as-promised"   and "sinon" are dev dependencies

PR #682, opened 2018-12-18
Updates to readme, adding library to sys path

devsatishm:
### Description

- Added `library` to system path
- Updated instructions on how to run unit tests in Python 2.X & Python 3.X environments


chimpuz:
Build works fine

PR #686, opened 2018-12-20
Features edit advance

glearner:
### Requirements

Updated the Service overview Component for advance edit mode. 

### Description of the Change

User will be able to edit advance fields of lambda functions.


supritat:
@glearner, I believe already slider replacement code is there in develop branch. Why does this PR includes the same changes?

PR #687, opened 2018-12-20
UI bug fix

zoexiong:
### Description of the Change

UI bug fix for go lang support, tested and verified based on today's new build

### Benefits
now all the options are in one line and console errors are gone

### DIFF
--- a/core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.html
+++ b/core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.html
@@ -98,8 +98,8 @@
                         </section>
                         <section class="col-lg-2 col-md-2 pad-left0 runtime-option">
                             <div class="radio-container">
-                                <input type="radio" name="runtime" id="go" [value]="go1.x" [checked]="runtime == 'go1.x'" (change)="onSelectionChange('go1.x')">
-                                <label for="go1.x"><span class='outer'></span><span class="background"></span>Go 1.x</label>
+                                <input type="radio" name="runtime" id="go" [value]="'go1.x'" [checked]="runtime == 'go1.x'" (change)="onSelectionChange('go1.x')">
+                                <label [for]="'go'"><span class='outer'></span><span class="background"></span>Go 1.x</label>
                             </div>
                         </section>
                     </div>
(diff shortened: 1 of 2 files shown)


core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.html
```diff
@@ -98,8 +98,8 @@ <h4 class="bold-title">
                         </section>
                         <section class="col-lg-2 col-md-2 pad-left0 runtime-option">
                             <div class="radio-container">
-                                <input type="radio" name="runtime" id="go" [value]="go1.x" [checked]="runtime == 'go1.x'" (change)="onSelectionChange('go1.x')">
-                                <label for="go1.x"><span class='outer'></span><span class="background"></span>Go 1.x</label>
+                                <input type="radio" name="runtime" id="go" [value]="'go1.x'" [checked]="runtime == 'go1.x'" (change)="onSelectionChange('go1.x')">
+                                <label [for]="'go'"><span class='outer'></span><span class="background"></span>Go 1.x</label>
```
devsatishm: Why are we reverting this change?

zoexiong: I just found out this option is not selectable without this change when I do the testing. Apologies that I didn't try to click the radio button yesterday when I test it. This time it's thoroughly tested.

devsatishm: A little confused though.  
How are other runtime options working? Is it something related to usage of period (version) in the string (`python` vs `go1.x`)?

zoexiong:
yes, it’s because of the period. the other options work fine.

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.html
supritat: @zoexiong, Tried fixing with your solution. HTML was parsing without any error but radio button wasn't working as expected.
I think attribute values should be assigned as following(when you are using period/version):
id="go1.x"
[value] = "'go1.x'"
[for]="'go1.x'"

PR #692, opened 2018-12-20
Email unit tests bug fix

svidhani:
### Requirements

Added missing package and fixed unit test


core/jazz_email/package.json
```diff
@@ -12,6 +12,7 @@
     "jshint": "jshint --exclude ./node_modules ."
   },
   "dependencies": {
+    "aws-sdk": "^2.380.0",
```
devsatishm: Why do we need this? In actual AWS environment, this is injected automatically (we don't need to package it during deployment explicitly). If you need this for running unit tests, you can add it to devDependencies.

svidhani: yes it was for unit tests. updated

PR #694, opened 2018-12-21
Disabling go runtime in the UI

devsatishm:
### Description of the Change

- Disabling `Go` runtime in the UI during service creation
- Getting backend services ready for `Go` integration


core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.html
```diff
@@ -98,7 +98,7 @@ <h4 class="bold-title">
                         </section>
                         <section class="runtime-option">
                             <div class="radio-container">
```
supritat: Include 'disable-event' class here to include css same as C#

devsatishm: Done

PR #697, opened 2018-12-24
Cognito Authorizer that validates user token & checks if user is an admin

supritat:
### Requirements

* Authorize requests so that only specific users (admin) are allowed to access the underlying API.

### Description of the Change

- New service: jazz_cognito-admin-authorizer
- Validates token
- If user=admin, creates auth policy so that API Gateway allows the request
- If not, creates auth policy so that API Gateway rejects the request


core/jazz_root-authorizer/index.js
```diff
@@ -0,0 +1,518 @@
+// =========================================================================
```
svidhani: better name will be jazz_admin-authorizer or jazz_cognito-admin-authorizer

supritat: Created new folder "jazz_cognito-admin-authorizer" and moved all the files.

core/jazz_root-authorizer/index.js
```diff
@@ -0,0 +1,518 @@
...
+  if (!pems) {
+    logger.info("NO PEM file");
+    //Download the JWKs and save it as PEM
+    request({
+      url: cognitoUserPoolEndpoint + '/.well-known/jwks.json',
+      json: true
+    }, function (error, response, body) {
+      if (!error && response.statusCode === 200) {
+        pems = {};
+        var keys = body.keys;
```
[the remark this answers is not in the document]
supritat: Worked on ES6 changes.

PR #698, opened 2018-12-24
Fixed logs

svidhani:
### Requirements

Removed the condition in log to change the levels

### Description of the Change

Not sure why the condition was set to not change the level of logging down to other levels but this change will allow that.

### DIFF
--- a/templates/api-template-nodejs/components/logger.js
+++ b/templates/api-template-nodejs/components/logger.js
@@ -89,21 +89,18 @@ module.exports = function () {
 
     const log = (level, message) => {
       const timestamp = new Date().toISOString();
-
-      if (logLevels[level] >= logLevels[config.curLogLevel]) {
-        if (level == 'error') {
-            console.error(timestamp, 'ERROR \t', config.requestDetails, message);
-        } else if (level == 'warn') {
-            console.warn(timestamp, 'WARN  \t', config.requestDetails, message);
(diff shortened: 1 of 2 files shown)


templates/api-template-nodejs/components/logger.js
```diff
@@ -89,21 +89,18 @@ module.exports = function () {
 
     const log = (level, message) => {
       const timestamp = new Date().toISOString();
-
-      if (logLevels[level] >= logLevels[config.curLogLevel]) {
```
svidhani: i don't know why this was there initially(this is what i had reverted to before my Es6 changes).

devsatishm: This is required to allow developers change log level during runtime (ex: from INFO to DEBUG) so that they see more logs. We select INFO as the default log level. But, if developers want better logging, they'll need to set log level to DEBUG/VERBOSE. Instead of doing this at code level (by updating https://github.com/tmobile/jazz/blob/develop/templates/lambda-template-nodejs/components/logger.js#L43), they can update the environment variable: LOG_LEVEL (=https://github.com/tmobile/jazz/blob/develop/templates/lambda-template-nodejs/components/logger.js#L77) so  that they don't need to deploy code again. 

Also, it allows us to enable deeper logging only when required and not all the time.

PR #710, opened 2018-12-27
:bug: Enabling event source mappings for Jazz event handlers by default + fixing auth token issue

devsatishm:
### Description of the Change

- Enabled event source mappings for Jazz event handlers by default 
- Fixed incorrect use of auth token in jenkins build packs

### Benefits

- Event source (Kinesis) mapping for all core handlers is enabled by default that will guarantee event processing.
- Create workflow will use the correct login token to call other core services.


builds/jazz-build-module/service-configuration-data-loader.groovy
```diff
@@ -532,7 +532,7 @@ def setEventSourceMapping(eventSourceArn, config) {
   ).trim()
   echo "$event_source_list"
   if (event_source_list == "[]") {
-      sh "aws lambda create-event-source-mapping --event-source-arn ${eventSourceArn} --function-name arn:aws:lambda:$region:$role_id:function:$function_name --starting-position LATEST --region $region"
+      sh "aws lambda create-event-source-mapping --event-source-arn ${eventSourceArn} --function-name arn:aws:lambda:$region:$role_id:function:$function_name --enabled --starting-position LATEST --region $region"
```
svidhani: this is interesting. I wonder how was it working before. Because the triggers were enabled which was not consistent though.

PR #714, opened 2019-01-01
Bug fix for jazz_cognito-admin-authorizer

supritat:
### Requirements

* Bug fix for jazz_cognito-admin-authorizer and jazz_create-serverless-service services.

### Description of the Change

* jazz_cognito-admin-authorizer: fixed constructor error that as arrow functions don't have this, they cannot be used as constructor.
* jazz_create-serverless-service: Passing authToken param to updateAclPolicy() and fix for error handling.

### Benefits

* jazz_cognito-admin-authorizer successfully validates the for admin authorization.
* jazz_create-serverless-service successfully send request to ACL service to update the policies for user created service.

### Possible Drawbacks

NA

### Applicable Issues

NA

### DIFF
--- a/core/jazz_cognito-admin-authorizer/index.js
+++ b/core/jazz_cognito-admin-authorizer/index.js
@@ -218,7 +218,7 @@ module.exports.handler = (event, context, cb) => {
    * @class AuthPolicy
    * @constructor
    */
-  var AuthPolicy = (principal, awsAccountId, apiOptions) => {
+  function AuthPolicy(principal, awsAccountId, apiOptions) {
     /**
      * The AWS account id the policy will be generated for. This is used to create
      * the method ARNs.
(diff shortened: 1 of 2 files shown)


core/jazz_cognito-admin-authorizer/index.js
```diff
@@ -402,7 +402,7 @@ module.exports.handler = (event, context, cb) => {
        *
        * @method allowAllMethods
        */
-      allowAllMethods: () => {
+      allowAllMethods: function () {
```
svidhani: why?

core/jazz_cognito-admin-authorizer/index.js
```diff
@@ -411,7 +411,7 @@ module.exports.handler = (event, context, cb) => {
        *
        * @method denyAllMethods
        */
-      denyAllMethods: () => {
+      denyAllMethods: function () {
```
svidhani: why?

supritat: It was throwing TypeError with arrow function. May be it is due to the context attached to "this".

supritat: It was throwing TypeError with arrow function. May be it is due to the context attached to "this".

core/jazz_cognito-admin-authorizer/index.js
supritat: It was throwing TypeError with arrow function. May be it is due to the context attached to "this".

PR #728, opened 2019-01-09
ACL Impl: Update GET /services and  /services/{id} 

supritat:
### Requirements

* As part of ACL feature implementation jazz_services API is updated to include policy details for respective services.

### Description of the Change

* GET /services - Filtering the list of services from DB based on the list of serviceIds sent from authorizer and attaching respective policies to the services.
* GET /services/{id} - Filtering the provided service from DB and attaching respective policies to the response.
* Swagger file - updated response object to include policies.

### Benefits

Services response will include service details along with policy details for the respective user.

### DIFF
--- a/core/jazz_services/components/crud/getList.js
+++ b/core/jazz_services/components/crud/getList.js
@@ -25,7 +25,7 @@
 const utils = require("../utils.js")();
 const _ = require("lodash");
 
-module.exports = (query, getAllRecords, onComplete) => {
+module.exports = (query, getAllRecords, servicesList, onComplete) => {
     // initialize dynamodb
     var dynamodb = utils.initDynamodb();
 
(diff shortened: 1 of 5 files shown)


core/jazz_services/components/validation/validate_create_payload.js
```diff
@@ -85,9 +85,8 @@ module.exports = (service_data, onComplete) => {
         },
         // Check if a service with same domain and service_name combination exists
         validateServiceExists: function (onComplete) {
-            getAllRecords = "true";
             var query = { service: service_data.service.toLowerCase(), domain: service_data.domain.toLowerCase() };
-            crud.getList(query, getAllRecords, function onServiceGet(error, data) {
+            crud.getList(query, null, function onServiceGet(error, data) {
```
svidhani: is this intended

supritat: This is not intended. It is not being used after the acl update, so removed getAllRecords.
getList() expect the list of serviceIds as param which will be null for POST request.

core/jazz_services/components/crud/getList.js
```diff
@@ -141,4 +154,18 @@ module.exports = (query, getAllRecords, onComplete) => {
         });
     };
     scanExecute(onComplete);
+
+    function appendPolicies(filteredList, servicesList) {
+        if (servicesList && servicesList.lengtht) {
```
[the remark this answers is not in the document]
supritat: Updated

PR #738, opened 2019-01-12
Fix bugs for adding and removing policies

svidhani:
### Requirements

Bug fixes for add and remove

### Description of the Change

Add policy when there is no policy for that service id in DB

### DIFF
--- a/core/jazz_acl/components/casbin.js
+++ b/core/jazz_acl/components/casbin.js
@@ -126,7 +126,7 @@ async function addOrRemovePolicy(serviceId, config, action, policies) {
               await enforcer.savePolicy();
             }
           } else {
-            result.error = `Rollback transaction - could delete ${removeResult.length} of ${totalPolices} policies`;
+            result.error = `Rollback transaction - could delete ${removeResult.length} of ${totalPolicies} policies`;
           }
         }
 


core/jazz_acl/components/casbin.js
```diff
@@ -98,19 +102,19 @@ async function addOrRemovePolicy(serviceId, config, action, policies) {
...
       let removeResult = [];
 
       if (totalPolicies) {
         conn = await dbConnection(config);
         const enforcer = await casbin.newEnforcer('./config/rbac_model.conf', conn);
 
-        // remove (incase of add remove first)
+        // remove (incase of add, remove first)
         if (action === 'remove' || action === 'add') {
           removeResult = await getPolicies.map(async policies => {
```
pspani: if totalpolicies is zero, then removal shouldn't happen i guess.

svidhani: yes and it wont(see line 113)

core/jazz_acl/components/casbin.js
```diff
@@ -98,19 +102,19 @@ async function addOrRemovePolicy(serviceId, config, action, policies) {
   let result = {};
   let conn;
   const objects = [`${serviceId}_manage`, `${serviceId}_code`, `${serviceId}_deploy`];
-  let totalPolices = 0;
+  let totalPolicies = 0;
```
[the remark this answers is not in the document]
svidhani: updated

PR #740, opened 2019-01-14
Bug fix to send the service detail even if policy details are not available from authorizer.

supritat:
### Requirements

* send the service detail even if policy details are not available from authorizer.

### Description of the Change

Included a condition in GET services/{id} method

### DIFF
--- a/core/jazz_services/index.js
+++ b/core/jazz_services/index.js
@@ -83,7 +83,10 @@ module.exports.handler = (event, context, cb) => {
         // 1: GET service by id (/services/{service_id})
         if (event.method === 'GET' && service_id) {
             logger.info('GET service by ID : ' + service_id);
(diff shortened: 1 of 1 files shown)


core/jazz_services/index.js
```diff
@@ -83,7 +83,10 @@ module.exports.handler = (event, context, cb) => {
         // 1: GET service by id (/services/{service_id})
         if (event.method === 'GET' && service_id) {
             logger.info('GET service by ID : ' + service_id);
-            let servicePermission = event.services[0].policies;
+            let servicePermission =[];
+            if (event.services && event.services.length) {
```
pspani: should we send service details if no policy details are available?  I am assuming no policy details means the user is not authorized for any services. is my assumption correct?

PR #744, opened 2019-01-15
Filter Name & Namespace with partial matches.

akhil-kn:
Advanced column-wise filtering for both "Name" & "Namespace" like global search.
### Description of the Change
Previously filter for both columns are based on the exact match of search pattern. Now returns the list with column value contains the search pattern.   

untitled1

untitled2

### Benefits

untitled3

### Possible Drawbacks

NA

### Applicable Issues

NA


akhil-kn:
> add/update relevant unit test case

Added.

supritat:
@akhil-kn, Please the resolve conflicts

supritat:
Changes look good. Merging PR

PR #746, opened 2019-01-15
:fire: removed third party libraries lodash

Shibon-ll:
### Requirements
Avoid using third party library for simple uses

### Description of the Change

Removed third party lodash library from dependencies

Example   removed _.included by core JS includes, _.find by core JS find, _.isEmpty by custom functions

### Benefits

Can avoid Maintenance headache 
Can avoid third party dependencies

### Possible Drawbacks

nill

### Applicable Issues

nill


core/jazz_cloud-logs-streamer/index.js
```diff
@@ -168,7 +167,7 @@ function transform(payload) {
                         data.message = message;
                     }
                     data.log_level = utils.getSubInfo(logEvent.message, globalConfig.PATTERNS.log_level, 0);
-                    if (_.isEmpty(data.log_level)) {
+                    if (isEmpty(data.log_level)) {
```
supritat: Instead of isEmpty function, you can write if(!data.log_level).

Shibon-ll: isEmpty function has been removed

PR #748, opened 2019-01-16
Bugfix/acl addremove

svidhani:
### Requirements

Bugfix for add remove policy

### DIFF
--- a/core/jazz_acl/components/casbin.js
+++ b/core/jazz_acl/components/casbin.js
@@ -109,11 +109,14 @@ async function addOrRemovePolicy(serviceId, config, action, policies) {
   try {
     const getPolicies = await getFilteredPolicy(1, objects, config);
 
+
     if (getPolicies && getPolicies.error) {//if there was any error capture that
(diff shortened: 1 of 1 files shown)


core/jazz_acl/components/casbin.js
```diff
@@ -111,9 +111,11 @@ async function addOrRemovePolicy(serviceId, config, action, policies) {
 
     if (getPolicies && getPolicies.error) {//if there was any error capture that
       result.error = getPolicies.error;
+      return result;
     }
-    else if (getPolicies && !getPolicies.error) {//found previous policies to be removed
-      totalPolicies = getPolicies.filter(policy => policy.length > 0).length;
+
+    totalPolicies = getPolicies ? getPolicies.filter(policy => policy.length > 0).length : 0;
+    if (getPolicies) {//found previous policies to be removed
```
SiniWilson: If getPolicies value is [[],[],[]], this will not work. Means line num 147 will not execute, which is supposed to execute.

PR #764, opened 2019-01-18
Error message while updating the services without any change in the existing data

gokulpits:
Error message while updating the services without any change in the existing data

### Description of the Change

It shows an error message like 'Something went wrong while updating the services'. when trying to edit the services without updating any of the existing data.

er1

### Benefits

Now the completed button enables only if there is any update in the existing value.
er3

### Possible Drawbacks

NA

### Applicable Issues

NA


suryajak:
@gokulpits can you fix the merge conflicts?

core/jazz_ui/src/app/pages/service-overview/service-overview.component.ts
```diff
@@ -463,7 +463,14 @@ export class ServiceOverviewComponent implements OnInit {
 
     }
     this.PutPayload = payload;
-    if (Object.keys(this.PutPayload).length > 0) this.isPayloadAvailable = true
+
+    if (Object.keys(this.PutPayload).length > 0) {
+      this.isPayloadAvailable = true
+    }
+    else {
```
SiniWilson: Avoid else by assigning  this.isPayloadAvailable = false   at first

PR #771, opened 2019-01-21
Deployments

svidhani:
### Requirements

Refine deployments


core/jazz_deployments-event-handler/test/test.js
```diff
@@ -622,8 +622,8 @@ describe("updateDeployments", () => {
     });
     index.updateDeployments(JSON.stringify(res), deploymentPayload, configData, "temp_auth").then((obj) => {
       sinon.assert.calledOnce(processRequestStub);
-      processRequestStub.restore();
     })
+    processRequestStub.restore();
```
pspani: why was this moved out from the callback

supritat: updated

PR #780, opened 2019-01-23
ACL - Unit tests

svidhani:
### Requirements

Unit tests for acl

### DIFF
--- a/core/jazz_acl/test/test.js
+++ b/core/jazz_acl/test/test.js
@@ -65,8 +65,31 @@ describe("Validation tests", () => {
       expect(() => validationUtil.validatePostPoliciesInput({body: {serviceId: '342342', policies: [{userId: '23222'}]}})).to.throw();
     });
   });
-});
 
+  describe("Validate get check perms input", () => {
+    it("validate throws error when userId is missing", () => {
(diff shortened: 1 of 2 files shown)


core/jazz_acl/test/test.js
```diff
@@ -65,8 +65,31 @@ describe("Validation tests", () => {
       expect(() => validationUtil.validatePostPoliciesInput({body: {serviceId: '342342', policies: [{userId: '23222'}]}})).to.throw();
     });
   });
-});
 
+  describe("Validate get check perms input", () => {
+    it("validate throws error when userId is missing", () => {
+      expect(() => validationUtil.validateGetCheckPermsInput({query: {}})).to.throw();
```
pspani: I think we should be asserting on the type of exception being thrown. Do you agree?

svidhani: the way this works(using throw) is without the type. Do you know a way to do it?

pspani: expect(() => validationUtil.validateGetCheckPermsInput({query: {}})).to.throw().with.property('message', 'User Id is missing'); may work.

PR #783, opened 2019-01-23
Delopyemts flow bug fix

supritat:
### Requirements

* Bug fix for refine deployments


builds/jazz-build-module/utility-loader.groovy
```diff
@@ -166,4 +166,9 @@ def getApiToken(){
 	}
 }
 
+def isBuildAReplay() {
```
SiniWilson: Please give some other name

builds/jazz-build-module/utility-loader.groovy
```diff
@@ -166,4 +166,9 @@ def getApiToken(){
...
+  def replyClassName = "org.jenkinsci.plugins.workflow.cps.replay.ReplayCause"
```
SiniWilson: replayClassName  instead ??

supritat: updated

builds/jazz-build-module/utility-loader.groovy
supritat: updated

PR #786, opened 2019-01-24
add azure function test for test_lambda

heina-chang:
### Requirements

* Currently AWS lambda is tested via lambda invoke which does not work for azure function
* Added a new way to test lambda API to test azure function via http invoke

### Description of the Change

Added a http call to very azure function

### Benefits

Support azure function testing from Jazz UI

### Possible Drawbacks

None, I added a separate flow for aws lambda vs azure function.  The existing lambda invoke stays the same.

### DIFF
--- a/core/jazz_test-lambda/components/utils.js
+++ b/core/jazz_test-lambda/components/utils.js
@@ -34,6 +34,16 @@ var validateARN = (arn) => {
   }
 
 };
+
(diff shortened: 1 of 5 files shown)


heina-chang:
close per Satish M.  this is a dups with https://github.com/tmobile/jazz/pull/841

PR #787, opened 2019-01-25
Integrating Jasmine Json Reporter Module

Johnbush:
### Requirements

* Preparing Detailed summary while sending UIAutomation Tests Results.

### Description of the Change

* Integrated Jasmine Json Reports Module 
* This will provide the test Results as Json Format which can be parsed to get the results in a clean way

### DIFF
--- a/core/jazz_ui/protractor.conf.js
+++ b/core/jazz_ui/protractor.conf.js
@@ -4,6 +4,7 @@
 const SpecReporter = require('jasmine-spec-reporter').SpecReporter;
 var Jasmine2HtmlReporter = require('protractor-jasmine2-html-reporter');
 const JasmineConsoleReporter = require('jasmine-console-reporter');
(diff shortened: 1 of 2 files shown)


core/jazz_ui/protractor.conf.js
```diff
@@ -4,6 +4,7 @@
 const SpecReporter = require('jasmine-spec-reporter').SpecReporter;
 var Jasmine2HtmlReporter = require('protractor-jasmine2-html-reporter');
 const JasmineConsoleReporter = require('jasmine-console-reporter');
+var JSONReporter = require('jasmine-json-test-reporter');
```
pspani: please add the package to package.json if necessary.

Johnbush: Added it on package json, but this is an import statement.

PR #826, opened 2019-02-11
Including token(simple) authorizer.

supritat:
### Requirements

* To validate authorization token.

### Description of the Change

Included jazz_cognito-token-authorizer service for simple auth token validation.

### Benefits

jazz_cognito-token-authorizer can be used for basic token validation.


core/jazz_cognito-token-authorizer/.jshintrc
```diff
@@ -0,0 +1,4 @@
+{
```
pspani: we dont usually check in jshintrc files. I would prefer we add jshintrc and edtior.config files as part of a story where we have discussed our coding standards. Till then lets delete this file and add the appropriate entry in .gitignore.

supritat: Removed .jshintrc file

Issue #625, opened 2018-11-17
Fix unit tests in logs microservice


suryajak:
### Description

A few unit tests are broken in the logs microservice (referenced in this forked repo: https://github.com/suryajak/jazz/commit/6b5afc0abd3ff168170c3c742f6ebe4a21833360) which will need to be fixed once this is merged to master.

### Steps to Reproduce

1. Uncomment the tests in the referenced commit and run the unit tests

**Expected behavior:** [What you expect to happen]
Tests should pass

**Actual behavior:** [What actually happens]
Tests fail 

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

davidsims9t:
Where is `npm test` defined? I don't see any package.json file.

davidsims9t:
I created a PR for this issue: https://github.com/suryajak/jazz/pull/2

suryajak:
Thanks @davidsims9t. The description is misleading in the sense that it is using a commit id to point out the issue but the commit id is from a forked repo. Can you raise a PR fixing this issue in 
https://github.com/tmobile/jazz/blob/master/core/jazz_logs/test/test.js?

Specifically the following unit tests needs to be uncommented:

1. https://github.com/tmobile/jazz/blob/master/core/jazz_logs/test/test.js#L255 - https://github.com/tmobile/jazz/blob/master/core/jazz_logs/test/test.js#L264 
2. https://github.com/tmobile/jazz/blob/master/core/jazz_logs/test/test.js#L271 - https://github.com/tmobile/jazz/blob/master/core/jazz_logs/test/test.js#L297
3. https://github.com/tmobile/jazz/blob/master/core/jazz_logs/test/test.js#L341 - https://github.com/tmobile/jazz/blob/master/core/jazz_logs/test/test.js#L373

Let me know if that makes sense.

davidsims9t:
@suryajak Yes I opened a PR against this repo. Also I noticed you didn't mention the test on lines 303 to 334. Should I ignore this test?

PR #830, opened 2019-02-13
Fix jazz logs unit test

davidsims9t:
### Requirements

* Run `npm test` in `core/jazz_logs`

### Description of the Change

I fixed the four broken unit tests in this branch as well as upgraded sinon to the latest version. I changed the request() call to request.post() so that I could stub this request in the unit test. request() apparently is impossible to stub which you can read about here: https://stackoverflow.com/questions/20050507/creating-request-stub-with-sinon-in-mocha

I replaced sinon.stub(object, "method", func) with sinon.stub(object, "method").callsFake(func) as the first method is deprecated.

I added LOG_LEVELS to the test config JSON file. This is required, otherwise map() blows up and makes the tests fail.

### Benefits

The unit tests pass

### Possible Drawbacks

I'm not sure if there's any potential issues by using request.post() instead of request(), but there shouldn't be.

### Applicable Issues

https://github.com/tmobile/jazz/issues/625


core/jazz_logs/test/test.js
```diff
@@ -6,6 +6,7 @@ const awsContext = require('aws-lambda-mock-context');
 const sinon = require('sinon');
 const logger = require("../components/logger.js");
 const request = require('request');
+const _ = require('lodash');
```
pspani: We are reducing our dependence on third-party libraries. One of them is lodash. We are replacing all lodash usages in the code with js native functions. Please remove lodash and use native functions

devsatishm:
Closing due to inactivity.

PR #832, opened 2019-02-14
Fix: Changed xpath for jazzservices test file

vijay4953:
Changed the old xpath as it was wrong written because of that it was unable to find the elements.


pspani:
raise the PR for develop branch

PR #838, opened 2019-02-15
Ensure that aws profiles are getting cleared/deleted after each build completion

rajeevr2715:
### Requirements

Ensure aws profiles are getting cleared/deleted after each build completion

### Description of the change

Currently we have issues with builds as the aws profiles are not cleared after build completion.

Possible reasons for the build issues:
1. We’ll need to clear the profile (reset keys) after every build.  It might mean that the change to use a new random profile name with every build had a bug (not clearing our profiles/keys).
2.  Clearing just the keys would still leave the profile (with dummy keys) in the creds file. This will make the file larger & larger. We’ll need to find a way to remove the profile  after the build instead of just clearing out the keys.

This change deletes the aws profile (references) after the builds gets completed (success/fail).

### Possible Drawbacks

NA

### Applicable Issues

NA

### DIFF
--- a/builds/delete-serverless-service-build-pack/Jenkinsfile
+++ b/builds/delete-serverless-service-build-pack/Jenkinsfile
@@ -849,7 +849,9 @@ def cleanupS3BucketPolicy(stage, assetInfo) {
 def resetCredentials(credsId) {
 	echo "resetting AWS credentials"
   def credPath = System.getenv().HOME + "/.aws/credentials"
-	sh "sed -i '/${credsId}/,+2d' ${credPath}"
+  def confPath = System.getenv().HOME + "/.aws/config"
+  sh "sed -i '/${credsId}/,+2d' ${credPath}"
+  sh "sed -i '/${credsId}/,+1d' ${confPath}"
 }
 @NonCPS
 def updateBucketPolicy(policy_json, bucketName){
--- a/builds/delete-serverless-service-build-pack/Jenkinsfile
+++ b/builds/delete-serverless-service-build-pack/Jenkinsfile
@@ -848,8 +848,6 @@ def cleanupS3BucketPolicy(stage, assetInfo) {
  */
 def resetCredentials(credsId) {
 	echo "resetting AWS credentials"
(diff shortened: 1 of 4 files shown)


rajeevr2715:
Done

rajeevr2715:
@SiniWilson cloud-api profile is not random, so removing this from file might cause issue in parallel builds.

builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -850,6 +850,8 @@ def resetCredentials(credsId) {
 	echo "resetting AWS credentials"
 	sh "aws configure set profile.${credsId}.aws_access_key_id XXXXXXXXXXXXXXXXXXXXXXXXXX"
 	sh "aws configure set profile.${credsId}.aws_secret_access_key XXXXXXXXXXXXXXXXXXXXXX"
```
pspani: if we are deleting the creds, then replacing the creds with XXXXX seems redundant

rajeevr2715: yeah, will remove

rajeevr2715: Done

builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -848,8 +848,8 @@ def cleanupS3BucketPolicy(stage, assetInfo) {
  */
 def resetCredentials(credsId) {
 	echo "resetting AWS credentials"
-	sh "aws configure set profile.${credsId}.aws_access_key_id XXXXXXXXXXXXXXXXXXXXXXXXXX"
-	sh "aws configure set profile.${credsId}.aws_secret_access_key XXXXXXXXXXXXXXXXXXXXXX"
+  def credPath = System.getenv().HOME + "/.aws/credentials"
+	sh "sed -i '/${credsId}/,+2d' ${credPath}"
```
devsatishm: @rajeevr2715  Can you check if we'll need to do something similar in  ~/.aws/config?

rajeevr2715: Yes. removed that also.

Issue #837, opened 2019-02-15
Python3.6 template error: AttributeError on 'exception_message = e.message'


deepusundar:
### Description
Python template 3.6 is not compatible with 3.6 run time. Look like the template is not tested for exception conditions. Python 3.x exception handling has been changed from 2.x version.   

### Steps to Reproduce
Raise an exception in the try block. The code fails with an 'AttributeError' on  e.message

pspani:
thanks @deepusundar for reporting this issue. This issue has been fixed by the team :)

PR #840, opened 2019-02-15
Fix Python3.6 Templates : Attribute issue while raising Exception

Johnbush:
### Requirements

Fixing Attribute error, while raising an exception in try block.

### Description of the Change

This occurs due to removal of **Exception.message** in Python3 versions. Fixed it by passing the  instance of exception **e** to get the custom message.


pspani:
fixes #837

PR #841, opened 2019-02-15
 Azure api, function and website features

heina-chang:
# Azure 

### Description of the Change

1. Enable Azure and persist platform to the service table as an optional fields for backward compatibility
2. Add azure metrics for storage
3. Add azure function tester in test-lambda repo
4. Add azure website creation and deployment
5. Add azure api creation and deployment
6. Add azure function creation and deployment for cron and events with language of node and C#
7. Add azure function event resources creation (new or existing resources Service Bus Queue, CosmosDB, Storage Blob, Event Hubs)
8. Add rollback when azure function creation failed
9. Add delete service to delete azure resources when click delete or delete branches
10. Add support to use existing resource group (this is backend only)

### Benefits

Support Azure API, Function and Website deployment to the azure cloud.
### Applicable Issues

Not yet support azure logging


heina-chang:
dups with #943

PR #845, opened 2019-02-19
Acl api update for super user scenario

supritat:
### Requirements

* Super user should have admin access to all the services.

### Description of the Change

* Updated functions for acl/services and acl/services/{id} resources
* If userId is super user, then get list of serviceIds from services catalog and append admin policies and return to authorizer.

### Benefits

Super user has admin permissions => Manage/admin, Code/write and Deploy/write.

### Possible Drawbacks

At present there is only one super user i.e., stack admin.

### DIFF
--- a/core/jazz_acl/components/casbin.js
+++ b/core/jazz_acl/components/casbin.js
@@ -18,7 +18,8 @@ const casbin = require('casbin');
 const TypeORMAdapter = require('typeorm-adapter');
 const logger = require('./logger.js');
 const errorHandlerModule = require("./error-handler.js")();
-const AWS = require("aws-sdk");
+const getList = require("./getList.js");
+const globalConfig = require("../config/global-config.json");
 
 /* Create a connection to the DB*/
 async function dbConnection(config) {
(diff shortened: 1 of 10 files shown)


core/jazz_acl/components/casbin.js
```diff
@@ -184,48 +189,130 @@ async function addPolicy(serviceId, policies, enforcer) {
...
+    userPolicies = userPolicies.map(policy => { return { permission: policy.permission, category: policy.category } });
+
+    return [{serviceId: serviceId, policies: userPolicies}];
+  }
+
+}
+
+/* fetch list of serviceIds from dynamodb */
+var scanResult;
+async function scanExecute(dynamodb, scanparams, items_formatted) {
```
SiniWilson: Do it in another file, since this file handles, casbin related stuffs

core/jazz_acl/components/casbin.js
```diff
@@ -184,48 +189,130 @@ async function addPolicy(serviceId, policies, enforcer) {
...
 
-  return [{serviceId: serviceId, policies: userPolicies}];
+  return scanResult
+}
+
+/* attach admin policies */
+function attachAdminPolicies(list) {
+  let svcIdList = []
+  list.forEach(eachId => {
+    let svcIdObj = {
```
SiniWilson: Read this from config  file, rather than hard coding here (Similar code already there)

core/jazz_acl/components/casbin.js
```diff
@@ -184,48 +189,130 @@ async function addPolicy(serviceId, policies, enforcer) {
...
-  let serviceIdSeen = new Set();
   let policies = [];
+  if (userId === config.SERVICE_USER) {
+    let items_formatted = [];
+    AWS.config.update({
+      region: "us-east-1"
+    });
 
-  if (result && result.error) {
+    let dynamodb = new AWS.DynamoDB({
```
SiniWilson: As like above do the dynamo db stuffs in another file

supritat: Updated

core/jazz_acl/components/casbin.js
supritat: Updated

core/jazz_acl/components/casbin.js
supritat: Updated

PR #848, opened 2019-02-21
Angular and  ReactJs website create selection 

gokulpits:
Angular and  ReactJs website create selections

### Description of the Change

Previously the website creation only through html but we impleted the website creations using angular and reactjs also.

-->

### Possible Drawbacks

NA

### Applicable Issues

NA


core/jazz_ui/src/environments/environment.oss.ts
```diff
@@ -6,6 +6,7 @@ export const environment = {
   envName: "oss",
   multi_env: {multi_env},
   slack_support: {slack_support},
+  WebLists: { "html" : "Html", "angular" : "Angular", "react": "ReactJS"},
```
chimpuz: First option should be "static".  Also lets maintain camelcase for the key "webLists".

gokulpits:
Suggestions changes completed

PR #852, opened 2019-02-21
Fix for super user request sent to access policies for particular service

supritat:
### Requirements

* Fetch super user policies for individual services.

### Description of the Change

* Check for the provided serviceId in services catalog and then return admin policies.

### Benefits

Super user policies for individual service are sent only if the serviceId is available in the services catalog.


core/jazz_acl/components/casbin.js
```diff
@@ -191,7 +191,11 @@ async function addPolicy(serviceId, policies, enforcer) {
 /* Get the permissions for a service given a userId */
 async function getPolicyForServiceUser(serviceId, userId, config) {
   if (userId === config.SERVICE_USER) {
-    let result = attachAdminPolicies([serviceId]);
+    const dbResult =  await getList.getSeviceIdList(config, serviceId)
+    if (dbResult && dbResult.error) {
```
pspani: !dbResult.error

supritat: This condition is taken care.

PR #860, opened 2019-02-26
Bug fix for endpoint and UI error module.

supritat:
### Requirements

* Bug fix for API endpoint and error module

### Description of the Change

* Environment logical ID will be part of API endpoint of AWS API services only.
* Bug fix for UI error module.


builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -315,7 +316,6 @@ node() {
               createSubscriptionFilters(stackName, configLoader.AWS.REGION, roleId);
             }
 
-            updateApiDocUrl(current_environment)
```
SiniWilson: Are you sure we dont need this method ??

supritat: we need it. Included in switch case.

PR #866, opened 2019-02-27
Update for default deployment target options.

supritat:
### Requirements

* Define default deployment targets for all the services.

### Description of the Change

* Default deployment target details are updated in config file
* For website services AWS_CLOUDFRONT will be the default deployment target.


core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.html
```diff
@@ -447,25 +447,6 @@ <h4 class="bold-title">
                 </div>
             </section>
             <!-- end of aws events -->
-            <section class="each-step-wrap cloudfront-url" *ngIf="typeOfService == 'website'">
```
devsatishm: I understand that this  feature is not enabled/used at this point, but let's keep it in here so that we can enable it if we need at a later point of time.

builds/jenkins-build-pack-website/Jenkinsfile
```diff
@@ -128,8 +128,8 @@ node()  {
 				}
 
 				create_cloudfront_url = true
-				if (service_config['create_cloudfront_url'] == true || service_config['create_cloudfront_url'] == "true") {
-					create_cloudfront_url = true
+				if (service_config['deployment_targets'] && service_config['deployment_targets'] instanceof Object && service_config['deployment_targets'].website == "aws_s3") {
```
devsatishm: Changes are good, let's target these for next release instead of targeting `apigee` release?

PR #871, opened 2019-02-28
Remove primary account role-policy

Johnbush:
### Requirements

While enabling multi-account feature we are creating couples of roles in non-primary account also we are updating the platform_services role of primary account to talk with cross-accounts. We are in need to drop the roles created by multi-account feature before stack destroy.

### Description of the Change

- updated with the deletion of platform_services  role on primary Account


pspani:
included in #883

PR #887, opened 2019-03-05
Acl implementation fix for services API and jazz_ui

supritat:
### Requirements

* Bug fix in services API to get empty service list if user hasn't created any service.
* jazz_ui bug fix: enable delete service/ edit buttons based on respective policies only.

### Description of the Change

* Services API: Return empty list if user hasn't created any service.
* jazz_ui: enabled delete service/ edit buttons based on respective policies.


SiniWilson:
Travis is failing. Please fix it

PR #913, opened 2019-03-13
Multi-Account Metrics API Cloudfront region fix

Johnbush:
### Description of the Change

- Updated the region as us-east-1, to which the AWS instance was created for accessing the cloudfront metrics.

### DIFF
--- a/builds/jazz-build-module/service-configuration-data-loader.groovy
+++ b/builds/jazz-build-module/service-configuration-data-loader.groovy
@@ -79,7 +79,7 @@ def loadServiceConfigurationData() {
             updateConfigValue("{jazz_admin}", config_loader.JAZZ.ADMIN);
             updateConfigValue("{jazz_admin_creds}", config_loader.JAZZ.PASSWD);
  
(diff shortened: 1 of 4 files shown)


builds/jazz-build-module/service-configuration-data-loader.groovy
```diff
@@ -58,7 +58,7 @@ def loadServiceConfigurationData() {
             updateConfigValue("{jazz_admin}", config_loader.JAZZ.ADMIN);
             updateConfigValue("{jazz_admin_creds}", config_loader.JAZZ.PASSWD);
  
-            sh "sed -i -- 's/{conf-region}/${region}/g' ./config/global-config.json"
+            sh "sed -i -- 's/{cf-region}/us-east-1/g' ./config/global-config.json"
```
pspani: add a comment on why was this done. Add the same in metrics api file too

Johnbush: updated.

PR #917, opened 2019-03-13
User management section in admin 

gokulpits:
User management section in admin side

### Description of the Change

There is no way to handle the users .

ub

### Benefits

Now we can handle the users from the admin side.From there we can enable and disable users also.

ua

### Possible Drawbacks

NA

### Applicable Issues

<NA


core/jazz_ui/src/app/core/services/admin-utils.service.ts
```diff
@@ -19,4 +19,14 @@ export class AdminUtilsService {
...
+    getAdminUsers(pageToken) {
+    console.log("token", pageToken)
+    let pageLimit = 10;
+    return this.http.get(`/jazz/usermanagement?limit=${pageLimit}&paginationtoken=${pageToken}`)
+      .toPromise()
+      .then((response) => {
+        if (response && response.data && response.data.users)
+          return response.data;
+      })
+  }
```
raghits: Check file formatting. Remove console.log.

SiniWilson:
@gokulpits  Please close this PR and raise a new one out of tmoble/jazz  ui-enhancements branch against the same, since some how this PR is rebased with sls-app branch and we could not review or validate your changes

PR #918, opened 2019-03-13
Multi-Account Delete build pack changes

Johnbush:
### Requirements

The Services created on non-primary accounts needs to be undeployed, while multi-account extension is being used on stack.

### Description of the Change

- Updated delete build to undeploy services hosted on cross Accounts.
- Undeployement of services are handled using the credentials of multiple Accounts stored under Jenkins.

### DIFF
--- a/builds/delete-serverless-service-build-pack/Jenkinsfile
+++ b/builds/delete-serverless-service-build-pack/Jenkinsfile
@@ -1,4 +1,4 @@
-#!groovy?
+#!groovy.
 import groovy.json.JsonOutput
 import groovy.transform.Field
 import groovy.json.JsonSlurperClassic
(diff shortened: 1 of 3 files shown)


builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -294,13 +298,9 @@ def archiveAssetDetails(env) {
  */
 def checkBucket(stage) {
 	withCredentials([
-		[$class: 'AmazonWebServicesCredentialsBinding', accessKeyVariable: 'AWS_ACCESS_KEY_ID', credentialsId: configLoader.AWS_CREDENTIAL_ID, secretKeyVariable: 'AWS_SECRET_ACCESS_KEY']
+		[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: service_config.credentialId]
 	]) {
 		try {
```
pspani: region would still be required to be set in environment variables. same in below too...

Johnbush: Regions are passed along with the cli commands.

PR #936, opened 2019-03-20
Updated test-lambda function (added permission in Lambda Jenkins)

SauarvD:
### Requirements

* Find a way to update jazz_test-lambda and give permission to platform services to invoke the function from the secondary account

### Description of the Change

Added permission to access the particular created function from primary account only if the deployment account is non-primary

### DIFF
--- a/builds/jenkins-build-pack-lambda/Jenkinsfile
+++ b/builds/jenkins-build-pack-lambda/Jenkinsfile
@@ -323,18 +323,15 @@ node() {
             createSubscriptionFilters(stackName, env_key, credsId, accountDetails);
           }
 
-          for (item in configLoader.AWS.ACCOUNTS) {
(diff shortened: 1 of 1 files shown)


builds/jenkins-build-pack-lambda/Jenkinsfile
```diff
@@ -323,6 +323,19 @@ node() {
             createSubscriptionFilters(stackName, env_key, credsId, accountDetails);
           }
 
+          for (item in configLoader.AWS.ACCOUNTS) {
+            if(item.ACCOUNTID == service_config.accountId){
+              if(!item.PRIMARY){
```
pspani: can we use accountDetails instead?

SauarvD: Used accountDetails

PR #948, opened 2019-03-25
Picking up the developer defined serverless.yml

olegfomindev9:
### Description of the Change

Picking up the developer defined serverless.yml and also verifying actions.


templates/sls-app-template-java/src/main/java/com/slf/util/EnvironmentConfig.java
```diff
@@ -33,18 +33,25 @@
 	public EnvironmentConfig (Context context) throws Exception {
 		String fnName = context.getFunctionName();
 		String fnNameNoEnv = null;
+		String fnNoPrefix = null;
 
 		if(null != fnName) {
 			int lastIndx = fnName.lastIndexOf("-");
 			stage = fnName.substring(lastIndx+1);
 			fnNameNoEnv = fnName.substring(0, lastIndx);
+			int preIndx = fnNameNoEnv.lastIndexOf("-");
```
raghits: Here we are assuming that the function name itself doesn't contain a '-' character. A lambda function can contain '-': https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#SSS-CreateFunction-request-FunctionName

Here fnNoPrefix is likely to get truncated at the '-'. So a function with name: "my-function", will probably just have fnNoPrefix: "function" instead of "my-function".

olegfomindev9: Ahhh, this a very good observation. Though honestly speaking I was continuing on the pattern of using the dash as a valid separator.

templates/sls-app-template-nodejs/function1/index.js
```diff
@@ -40,6 +40,7 @@ module.exports.handler = (event, context, cb) => {
 
   } catch (e) {
     //Sample Error response for internal server error
+		console.log('----e------>', e);
```
SiniWilson: Please remove this console

PR #965, opened 2019-04-09
[Fix] environment and updated with new framework structure

vijay4953:
updated with new framework structure that includes common functionality and also fix the issue of environment module.


vijay4953:
update the same changes to this PR https://github.com/tmobile/jazz/pull/870

PR #972, opened 2019-04-12
[WIP]Serverless Application Template for Python

Johnbush:
### Requirements

Server less Apps can contain multiple functions, where the existing templates support only one function.
Templates should refactored to support multiple functions.

### Description of the Change

Multiple functions can be created under folder named **functions**. Each function should have their own handler function and their respective test scripts and config files. Error, Logger, Config Modules are shared along the templates and this should be imported on each functions.


olegfomindev9:
Right now I can see only one function that is functions/func1 and the unit test does not test that function either :-(

Johnbush:
Closing this PR and creating new one https://github.com/tmobile/jazz/pull/979

PR #979, opened 2019-04-17
Serverless Application Template for Python

Johnbush:
### Requirements

This was the followup PR created for PR- https://github.com/tmobile/jazz/pull/972
Server less Apps can contain multiple functions, where the existing templates support only one function.
Templates should refactored to support multiple functions.

### Description of the Change

Multiple functions can be created under folder named functions. Each function should have their own handler function and their respective test scripts and config files. Error, Logger, Config Modules are shared along the templates and this should be imported on each functions.


olegfomindev9:
Two functions and meaningful unit tests. I like it :-)

PR #991, opened 2019-04-30
Including asset_type filter for metrics API

supritat:
### Requirements

* Metrics should be sent based on asset_type filter

### Description of the Change

Updated metrics.json with metricName and dimensions for other assets.

### Benefits

Metrics will be fetched for individual asset of respective service.


core/jazz_metrics/components/metrics.json
```diff
@@ -123,132 +166,298 @@
...
+            "Unit": "Milliseconds"
+          },
+          {
+            "Label": "Request Processing Latency",
+            "MetricName": "request_processing_latency",
+            "Statistics": "Min",
+            "Unit": "Milliseconds"
+          }
+        ],
+        "dimensions": ["apiproxy"]
```
raghits: Can we fix these whitespace changes?

supritat: After applying beautify Whitespace got added.

supritat: After applying beautify Whitespace got added.

raghits: But that just messes the code annotation/blame history. If it is not too much effort, can u remove those manually? Maybe Sini had similar issues.

supritat: Done

PR #1000, opened 2019-05-07
Fix/skipTest

vijay4953:
## Requirements

Catch errors and fail the test case when create-service fails.

## Description of the Change

Used flag and pending method to skip test when create-service fails.

## Benefits

It will help to find where exactly the test case fails and execute the other test cases.

## Possible Drawbacks
Need to handle some other elements too if anything found unexpected because of this change.

## Applicable Issues
NA


core/jazz_ui/e2e/common/commontest.ts
```diff
@@ -17,42 +17,52 @@ import { CONFIGURATIONS } from '../../src/config/configuration';
...
+let login_po: Login;
+login_po = new Login();
+
+
+export class Common {
+  public static readonly emailId = CONFIGURATIONS.optional.general.e2e.EMAIL_ID;
+  public static readonly config = CONFIGURATIONS.optional.general.e2e;
+  public static readonly timeOutHigh = 180000;
+  public static readonly EC = protractor.ExpectedConditions;
+  public static readonly microWait = 2000;
```
raghits: This makes sense!

PR #1032, opened 2019-05-28
Adding default value to the user provided non primary rules

SiniWilson:
### Requirements

Adding default value to the non primary rules

### Description of the Change

Adding default value to the user provided non primary rules
Fixing issues of develop sync with sls-app branch

### Benefits

Templates will pick the default value if the user does not specify any value for mandatory fields


builds/jazz-build-module/sls-app/sbr.groovy
```diff
@@ -673,7 +683,8 @@ class SBR_To_Constraint implements SBR_Constraint {
   }
 
   public boolean compliant(val) {
-    return val <= toValue;
+    if(val && val !== '')  return Integer.parseInt(val.toString()) <= Integer.parseInt(toValue.toString());
```
raghits: Do we need to handle ParseException?

builds/jazz-build-module/sls-app/sbr.groovy
```diff
@@ -685,7 +696,8 @@ class SBR_From_Constraint implements SBR_Constraint {
   }
 
   public boolean compliant(val) {
-     return val >= toValue;
+    if(val && val !== '')  return Integer.parseInt(val.toString()) >= Integer.parseInt(fromValue.toString());
```
raghits: Do we need to handle ParseException?

builds/jazz-build-module/sls-app/sbr.groovy
SiniWilson: Added try catch block

builds/jazz-build-module/sls-app/sbr.groovy
SiniWilson: Added try catch block

PR #1052, opened 2019-06-07
Read Deployment Descriptor from Environments table

Johnbush:
### Requirements

* when trying to build a particular environment after modifiying the deployment descriptor of that environment, the build pack will not get the updated deployment descriptor.

### Description of the Change

- Get deployment descriptor from the enviroments table.

### DIFF
--- a/builds/jazz-build-module/environment-deployment-metadata-loader.groovy
+++ b/builds/jazz-build-module/environment-deployment-metadata-loader.groovy
@@ -96,6 +96,22 @@ def getEnvironmentInfo() {
 	}
 }
 
+def getEnvDeploymentDecscriptor() {
+    def envdeploymentdescriptor =null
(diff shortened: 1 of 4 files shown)


builds/jazz-build-module/environment-deployment-metadata-loader.groovy
```diff
@@ -96,22 +96,27 @@ def getEnvironmentInfo() {
 	}
 }
 
-def getEnvDeploymentDecscriptor() {
+def getEnvDeploymentDecscriptor(env_logical_id) {
```
SiniWilson: No need to pass  the env_logical_id here, instead we can use the g_environment_logical_id

Johnbush: updated.

PR #1055, opened 2019-06-10
Fixing mandatory field wildcard isssue

SiniWilson:
### Requirements

Fixing mandatory field wildcard isssue

### Description of the Change

Fixed the role issue
Fixed the wildcard issue for the mandatory fields

### Benefits

User can able to specify the role in his deployment descriptor
user can able to give more complex structures in his deployment descriptor

### DIFF
--- a/builds/jazz-build-module/sls-app/serverless-build-rules.yml
+++ b/builds/jazz-build-module/sls-app/serverless-build-rules.yml
@@ -152,8 +152,8 @@ provider:
     sbr-type: int
     sbr-render: user-wins
     sbr-constraint:
-        sbr-from: 100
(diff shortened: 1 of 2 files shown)


builds/jazz-build-module/sls-app/serverless-build-rules.yml
```diff
@@ -152,11 +152,11 @@ provider:
     sbr-type: int
     sbr-render: user-wins
     sbr-constraint:
-        sbr-from: 5
-        sbr-to: 13
+        sbr-from: 100
```
raghits: can u fix the spacing issue.. please use spaces instead of tabs

PR #1065, opened 2019-06-12
Updates to deployment descriptor Layout

swati-sh:
Requirements
* First few lines of deployment descriptor should not be edited by user.
 Description of the Change
* First few lines should not be edited by user and included line number to deployment descriptor


raghits:
@swati-sh  Has this been applied to develop stack?

swati-sh:
Yes I have applied the changes to current stack for create service

PR #1090, opened 2019-06-27
Deployment Descriptor layout changes on environment-overview

swati-sh:
Description of the Change

First 8 lines is non-editable of deployment descriptor on environment-overview.
Condition for both function template and start new deployment descriptor.


raghits:
Couldn't validate on stack. Though Swati was able to. Will re-check.

PR #1115, opened 2019-07-10
Fixing jazz ui build issue

SiniWilson:
### Requirements

🐛 Jazz ui has to trigger the build for a new commit or new branch creation

### Description of the Change

Added  jazz ui as part of the catalog


raghits:
@SiniWilson Can we go over this tonight?

PR #1116, opened 2019-07-11
Cleanup jenkins workspace

rajeevr2715:
### Requirements

* We are using Fargate as the Container Service for our ECS to host gitlab and jenkins in our training environment. Fargate has a 10 GB limit on the disk space which we are going over almost immediately after we setup Jazz. We need to delete the build folders to free up diskspace. See: https://github.com/tmobile/jazz-installer/wiki#limitations

### Description of the Change

Added jenkins workspace cleanup method

### Benefits

Avoid "No Disk Space left" error due to disk size limit in ECS

### Possible Drawbacks

NA

### Applicable Issues

NA


raghits:
@rajeevr2715 How can we validate this?

PR #1121, opened 2019-07-24
Fixes metrics filter

swati-sh:
Description of the Change

Now asset type is on the top and asset name is at the bottom for sls-app


raghits:
Noted some issues - jumping and Asset name should just be Function1, Function2: 
This is when it is loading... 
image

This is when it is loaded
image

PR #1135, opened 2019-08-21
Disable tabs

swati-sh:
Description of the Change
  * Disable other tabs until overview is loaded

### DIFF
--- a/core/jazz_ui/src/app/pages/service-detail/service-detail.component.ts
+++ b/core/jazz_ui/src/app/pages/service-detail/service-detail.component.ts
@@ -45,7 +45,7 @@ export class ServiceDetailComponent implements OnInit {
 
   @Output() deleteServiceStatus: EventEmitter<boolean> = new EventEmitter<boolean>();
   @ViewChild('selectedTabComponent') selectedTabComponent;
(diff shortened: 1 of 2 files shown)


core/jazz_ui/src/app/pages/service-detail/service-detail.component.ts
```diff
@@ -45,7 +45,7 @@ export class ServiceDetailComponent implements OnInit {
 
   @Output() deleteServiceStatus: EventEmitter<boolean> = new EventEmitter<boolean>();
   @ViewChild('selectedTabComponent') selectedTabComponent;
-  isENVavailable: boolean = true;
+  isENVavailable: boolean = false;
```
raghits: Can we name it as isEnvAvailable so it is more readable? Also, before adding new variables make sure we do NEED this additional variable? Let us use camelCase the variable names going forward.

swati-sh: Updated the variable to camelCase. This variable was already available ,I just use it.

raghits: LGTM

PR #1157, opened 2019-09-23
API for managing CRUD of users/safes/role with t-vault

SiniWilson:
### Requirements

API for managing CRUD of users/safes/role with t-vault

### Description of the Change

- To add or remove  Safe on service creation and deletion 
- To add or remove or get roles for a given safe 
- To create user in Tvault with default policy  
- To add or remove users from Safes


core/jazz_t-vault/swagger/swagger.json
```diff
@@ -0,0 +1,3589 @@
+{
+    "swagger": "2.0",
+    "info": {
+        "version": "2017-01-30T23:30:18Z",
```
devsatishm: version: 1.0.0

core/jazz_t-vault/swagger/swagger.json
```diff
@@ -0,0 +1,3589 @@
...
+        "title": "{api_deployment_node_title}",
+        "description": "API specification for jazz t-valut api"
```
devsatishm: API specification for jazz t-valut api => Specification for APIs to manage secrets using Jazz & T-Vault

core/jazz_t-vault/swagger/swagger.json
```diff
@@ -0,0 +1,3589 @@
...
+        "/{service_name}/safes": {
+            "options": {
+                "description": "CORS Support",
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "tags": [
```
devsatishm: Let's use the following value for tags for each API/method - 
"tags" : ["Secret Management"]

core/jazz_t-vault/swagger/swagger.json
```diff
@@ -0,0 +1,3589 @@
...
+                ],
+                "consumes": [
+                    "application/json"
+                ],
+                "produces": [
+                    "application/json"
+                ],
+                "security": [
+                    {
+                        "{envPrefix}-{domain}-cognito-authorizer": []
```
devsatishm: Only `/safes/{safeName}/role` API should use this authorizer. All the remaining APIs - creation/deletion of safes/users, adding/removing users to the safe is done by admin user (hence uses `{envPrefix}-{domain}-cognito-admin-authorizer`)

core/jazz_t-vault/swagger/swagger.json
```diff
@@ -0,0 +1,3589 @@
...
+                    },
+                    "contentHandling": "CONVERT_TO_TEXT",
+                    "type": "aws",
+                    "responses": {
+                        ".*\"InternalServerError\".*": {
+                            "statusCode": "500",
+                            "responseTemplates": {
+
```
devsatishm: Model starts with CAPS and skip using Obj at the end wherever applicable. This can be `RoleDetails`.

devsatishm: Does this method return details for all the roles assigned to a safe? Or, is it for one getting details for one role for a particular safe? 

If this for one role, do we have an API to get list of all roles (with details) assigned to a safe?

devsatishm: We are not really creating roles. We are assigning roles to a safe. Update the model name to `AssignRoleToSafe`.

devsatishm: I would prefer keeping this empty instead of generic descriptions like these.

devsatishm: Most of these model definitions looks exactly the same. Can we use one definition and use that if  they don't differ at all?

devsatishm: Incorrect description, we are removing a role from safe..

devsatishm: Incorrect description

devsatishm: Incorrect description.

devsatishm: Incorrect description.

devsatishm: Incorrect description, we are adding a user to a safe.

devsatishm: Can we define an enum for the access/accessType?

devsatishm: Incorrect description.

devsatishm: Do we need to provide complete user details to remove user access to the safe? `username` should be good enough, right?

devsatishm: I think username should be good enough to remove user from safe.

devsatishm: Need more information here. Why is policies a string? Is this a list of permissions?

devsatishm: We'll just need username to remove/delete a user from vault.

devsatishm: Fix the description, prefer `Removes user from T-Vault` instead.

devsatishm: Every object has input in the response. Why is this different from others?

devsatishm: Use `Safe` instead.

devsatishm: `name`

devsatishm: UpdateSafeObj

devsatishm: Fix description

devsatishm: UserPermissions

devsatishm: Need more description. Enum type preferred is this can come from a known list of values.

devsatishm: I like this. Can we also add roles to this? It represents a safe along with all the users/roles and their permissions.

devsatishm: Fix description, we are not creating roles, we will only add/assign roles to the safes.

devsatishm: Will `User` suffice?

devsatishm: `username` should be fine.

core/jazz_t-vault/config/dev-config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "T_VAULT_API": "{tvault-host-name}/vault/v2",
+  "T_VAULT_USERNAME": "{tvault-username}",
+  "T_VAULT_PASSWORD": "{tvault-password}",
+  "T_VAULT_LOGIN_API": "/auth/tvault/login",
```
devsatishm: Some of  these paths are generic (environment agnostic), move them to global-config.json

core/jazz_t-vault/config/dev-config.json
```diff
@@ -0,0 +1,14 @@
...
+  "SAFE_API": "/ss/sdb",
+  "USER_IN_SAFE_API": "/user",
+  "ROLE_IN_SAFE_API": "/role",
+  "DELETE_SAFE": "/delete?path=shared/",
+  "GET_SAFE_DETAILS": "?path=shared/",
+  "GET_ROLE_DETAILS": "/auth/aws/role/",
+  "CREATE_ROLE_IN_VAULT": "/ss/auth/aws/iam/role?path=shared/",
+  "USER_IN_VAULT": "/auth/userpass/users"
```
devsatishm: Move these under a key (`API`) to keep them separate from other configurations.

```
{
    "API": {
        "SAFE": "/ss/sdb",
        "SAFE_USERS": "/user",
        "SAFE_ROLES": "/role",
        "DELETE_SAFE": "/delete?path=shared/",
        "GET_SAFE": "?path=shared/",
        "GET_ROLE": "/auth/aws/role/",
        "CREATE_ROLE": "/ss/auth/aws/iam/role?path=shared/",
        "USERS": "/auth/userpass/users"
    }
}
```

core/jazz_t-vault/config/global-config.json
```diff
@@ -0,0 +1,21 @@
+{
+  "CREATE_SAFE_REQUIRED_FIELDS": [
+    "safeName",
```
devsatishm: update these as described in swagger - 
`safeName` => `name`
`userName` => `username`
`roleName` => `name`

core/jazz_t-vault/config/dev-config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "T_VAULT_API": "{tvault-host-name}/vault/v2",
```
devsatishm: Use `T-VAULT` instead of `T_VAULT` (represents that it's part of the same word).

core/jazz_t-vault/index.js
```diff
@@ -0,0 +1,499 @@
...
+      } else {
+        return resolve(data);
+      }
+    });
+  });
+}
+
+const exportable = {
+  handler,
+  validateCreateSafeInput,
```
devsatishm: Move all the validation related functions to `utils/validations.js`?

core/jazz_t-vault/components/util.js
```diff
@@ -0,0 +1,387 @@
+// =========================================================================
```
devsatishm: I would rename this to `utils/vault.js`?

core/jazz_t-vault/index.js
```diff
@@ -0,0 +1,499 @@
...
+global.globalConfig = globalConfig;
+
+function handler(event, context, cb) {
+  logger.init(event, context);
+
+  try {
+    const configData = configModule.getConfig(event, context);
+    const resourcePath = event.resourcePath;
+
+    logger.info ("resourcePath : " + resourcePath);
```
devsatishm: debug.

core/jazz_t-vault/index.js
```diff
@@ -0,0 +1,499 @@
...
+          return cb(null, responseObj(result, event.body));
+        })
+        .catch(err => {
+          logger.error("Error creating safes: " + JSON.stringify(err));
+          if (err.errorType && err.errorType === "inputError") return cb(JSON.stringify(errorHandler.throwInputValidationError(err.message)));
+          else return cb(JSON.stringify(errorHandler.throwInternalServerError(err.error)));
+        });
+    }
+
+    if (event && event.method === 'GET' && resourcePath.endsWith("/safes/{safeName}")) {
```
devsatishm: Are you sure that the `resourcePath` that you get from event contains /safes/{safeName} even when request is `/safes/foobarsafe`?

core/jazz_t-vault/index.js
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    //SAFE
+    if (event && event.method === 'POST' && resourcePath.endsWith("/safes")) {
+      exportable.validateCreateSafeInput(event)
+        .then(() => { return exportable.genericInputValidation(event.body) })
+        .then(() => { return utils.getVaultToken(configData) })
+        .then((vaultToken) => { return exportable.createSafe(event.body, configData, vaultToken) })
+        .then((result) => {
+          logger.info("Successfully created safes. ");
```
devsatishm: safes => safe

core/jazz_t-vault/components/util.js
```diff
@@ -0,0 +1,387 @@
...
+      json: {
+        "username": configData.T_VAULT_USERNAME,
+        "password": configData.T_VAULT_PASSWORD
+      }
+    };
+
+    logger.debug("login payload : " + JSON.stringify(payload));
+    request(payload, function (error, response, body) {
+      logger.debug("login response : " + JSON.stringify(response));
+      if (response.statusCode && (response.statusCode === 200 || response.statusCode === 201) && body && body.client_token) {
```
devsatishm: Why is response code 200 or 201 for the same API?

core/jazz_t-vault/components/util.js
```diff
@@ -0,0 +1,387 @@
...
+function createSafe(safeDetails, configData, vaultToken, onComplete) {
+
+  let payload = {
+    uri: `${configData.T_VAULT_API}${configData.SAFE_API}`,
+    method: "POST",
+    headers: {
+      "vault-token": vaultToken
+    },
+    json: {
+      "path": `shared/${safeDetails.safeName.toLowerCase()}`,
```
devsatishm: Pull this (`/shared`) from config.

core/jazz_t-vault/components/util.js
```diff
@@ -0,0 +1,387 @@
...
+  logger.debug("createSafe payload : " + JSON.stringify(payload));
+  request(payload, function (error, response, body) {
+    logger.debug("createSafe response : " + JSON.stringify(response));
+    if (response.statusCode && (response.statusCode === 200 || response.statusCode === 201) && body) {
+      logger.debug("Successfully created safe: ");
+      return onComplete(null, body);
+    } else {
+      logger.error("Error in creating safe. " + JSON.stringify(response));
+      return onComplete({
+        "error": "Error in creating safe. " + response.body.errors
```
devsatishm: Let's also log the http error along with vault errors.

core/jazz_t-vault/index.js
devsatishm: Can we use `if .. else..` for these blocks of code?

core/jazz_t-vault/config/dev-config.json
raghits: @devsatishm `-` in key of map is not considered good practice as it is can't be directly referenced eg. `map.T_VAULT`. Would you prefer `TVAULT_API` or `T_VAULT_API`?

core/jazz_t-vault/swagger/swagger.json
SiniWilson: /{safeName} - GET will return all the details associated with the safe including role.  This particular method will return the details of the role which we are passing as query param.
And we dont have a separate method for getting all the roles of the particular safe.

SiniWilson: Taken data as common and not input since it is varying

SiniWilson: The UserDetails (now renamed) object contains only username. Since it is referring multiple times, defined as reference.

SiniWilson: Removed it from swagger, as it is predefined.

SiniWilson: The UserDetails (now renamed) object contains only username. Since it is referring multiple times, defined as reference.

SiniWilson: Removed from the swagger, since it is predefined. And yes, it is a string not an array/list

SiniWilson: Since user object is there in safe and vault, in order to distinguish the objects we may need this.

core/jazz_t-vault/config/global-config.json
SiniWilson: Renamed safeName as name in the request/response payloads and related fields. All other places renamed to safename. roleName renamed to rolename, as it is referenced in the same place where the safe name also used.

core/jazz_t-vault/index.js
SiniWilson: Done

core/jazz_t-vault/components/util.js
SiniWilson: Done

core/jazz_t-vault/components/util.js
SiniWilson: Done

core/jazz_t-vault/components/util.js
SiniWilson: Done

core/jazz_t-vault/components/util.js
```diff
@@ -0,0 +1,387 @@
...
+      },
+      json: {
+        "username": configData.T_VAULT_USERNAME,
+        "password": configData.T_VAULT_PASSWORD
+      }
+    };
+
+    logger.debug("login payload : " + JSON.stringify(payload));
+    request(payload, function (error, response, body) {
+      logger.debug("login response : " + JSON.stringify(response));
```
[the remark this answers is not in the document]
SiniWilson: Done

core/jazz_t-vault/components/util.js
```diff
@@ -0,0 +1,387 @@
...
+        "Content-Type": "application/json",
+        "Accept": "application/json"
+      },
+      json: {
+        "username": configData.T_VAULT_USERNAME,
+        "password": configData.T_VAULT_PASSWORD
+      }
+    };
+
+    logger.debug("login payload : " + JSON.stringify(payload));
```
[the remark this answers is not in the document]
SiniWilson: Done

core/jazz_t-vault/index.js
SiniWilson: Done

core/jazz_t-vault/index.js
SiniWilson: Done

core/jazz_t-vault/components/util.js
SiniWilson: Done

core/jazz_t-vault/index.js
SiniWilson: Done

core/jazz_t-vault/config/dev-config.json
SiniWilson: Done

core/jazz_t-vault/config/dev-config.json
SiniWilson: Done

core/jazz_t-vault/swagger/swagger.json
SiniWilson: Done

SiniWilson: Done

SiniWilson: I had added roles here. In the actual tvault API response we are getting "aws-roles" along with the safe details. But in aws-role, since there is a hyphen, I am getting errors while deploying swagger. So I had changed to "roles". Please advise

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

SiniWilson: Done

core/jazz_t-vault/swagger/swagger.json
SiniWilson: Done

core/jazz_t-vault/swagger/swagger.json
SiniWilson: Done

core/jazz_t-vault/swagger/swagger.json
SiniWilson: Done

core/jazz_t-vault/swagger/swagger.json
SiniWilson: Done

core/jazz_t-vault/swagger/swagger.json
```diff
@@ -0,0 +1,3614 @@
...
+                        "application/x-www-form-urlencoded": "\n#define( $body )\n  {\n  #foreach( $token in $input.path('$').split('&') )\n    #set( $keyVal = $token.split('=') )\n    #set( $keyValSize = $keyVal.size() )\n    #if( $keyValSize >= 1 )\n      #set( $key = $util.urlDecode($keyVal[0]) )\n      #if( $keyValSize >= 2 )\n        #set( $val = $util.urlDecode($keyVal[1]) )\n      #else\n        #set( $val = '' )\n      #end\n      \"$key\": \"$val\"#if($foreach.hasNext),#end\n    #end\n  #end\n  }\n#end\n\n#define( $loop )\n  {\n  #foreach($key in $map.keySet())\n      \"$util.escapeJavaScript($key)\":\n        \"$util.escapeJavaScript($map.get($key))\"\n        #if( $foreach.hasNext ) , #end\n  #end\n  }\n#end\n\n{\n  \"body\": $body,\n  \"method\": \"$context.httpMethod\",\n  \"principalId\": \"$context.authorizer.principalId\",\n  \"stage\": \"$context.stage\",\n\n  #set( $map = $input.params().header )\n  \"headers\": $loop,\n\n  #set( $map = $input.params().querystring )\n  \"query\": $loop,\n\n  #set( $map = $input.params().path )\n  \"path\": $loop,\n\n  #set( $map = $context.identity )\n  \"identity
```
devsatishm: This is incorrect use of `allOf` (you don't want `items` key in this object, do you?). Can you load this into https://editor.swagger.io/ & fix the swagger?

I see this usage across other objects too.

devsatishm: `UserDetails` & `UserDetailsForVault` look the same - contains just `username` & `password`.

raghits: It is easier to visualize and fix the issues with the swagger once loaded in the swagger editor.

SiniWilson: Done

SiniWilson: UserDetails - contains just username and UserDetailsForVault  contains both username and password

core/jazz_t-vault/swagger/swagger.json
```diff
@@ -0,0 +1,3608 @@
...
+                        "application/x-www-form-urlencoded": "\n#define( $body )\n  {\n  #foreach( $token in $input.path('$').split('&') )\n    #set( $keyVal = $token.split('=') )\n    #set( $keyValSize = $keyVal.size() )\n    #if( $keyValSize >= 1 )\n      #set( $key = $util.urlDecode($keyVal[0]) )\n      #if( $keyValSize >= 2 )\n        #set( $val = $util.urlDecode($keyVal[1]) )\n      #else\n        #set( $val = '' )\n      #end\n      \"$key\": \"$val\"#if($foreach.hasNext),#end\n    #end\n  #end\n  }\n#end\n\n#define( $loop )\n  {\n  #foreach($key in $map.keySet())\n      \"$util.escapeJavaScript($key)\":\n        \"$util.escapeJavaScript($map.get($key))\"\n        #if( $foreach.hasNext ) , #end\n  #end\n  }\n#end\n\n{\n  \"body\": $body,\n  \"method\": \"$context.httpMethod\",\n  \"principalId\": \"$context.authorizer.principalId\",\n  \"stage\": \"$context.stage\",\n\n  #set( $map = $input.params().header )\n  \"headers\": $loop,\n\n  #set( $map = $input.params().querystring )\n  \"query\": $loop,\n\n  #set( $map = $input.params().path )\n  \"path\": $loop,\n\n  #set( $map = $context.identity )\n  \"identity\": $loop,\n\n  #se
```
devsatishm: Shouldn't this be an array (of users?)

devsatishm: Shouldn't this be an array (of roles?)

core/jazz_t-vault/components/utils/vault.js
```diff
@@ -0,0 +1,389 @@
...
+    rejectUnauthorized: false
+  };
+
+  logger.debug("getSafeDetails payload : " + JSON.stringify(payload));
+  request(payload, function (error, response, body) {
+    logger.debug("getSafeDetails response: " + JSON.stringify(response));
+    if (response.statusCode && response.statusCode === 200) {
+      const safeDetails = JSON.parse(body);
+      logger.debug("Successfully got safe details");
+      return onComplete(null, safeDetails.data);
```
devsatishm: I see that GET /safes/{safeName} returns safe details along with list of users & roles and corresponding permissions. However, I don't see any specific calls to vault getting those details or logic to massage the data coming from vault. 

Can you check if we are returning data as per the swagger spec?

core/jazz_t-vault/index.js
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    else if (event && event.method === 'DELETE' && resourcePath.endsWith("/safes/{safename}")) {
+      validations.validateSafeInput(event)
+        .then(() => { return validations.genericInputValidation(event.path) })
+        .then(() => { return vault.getVaultToken(configData) })
+        .then((vaultToken) => { return exportable.deleteSafe(event.path.safename, configData, vaultToken) })
+        .then(result => {
+          logger.info("Successfully deleted safe");
+          return cb(null, responseObj(result, event.path));
```
devsatishm: Why are we returning `event.path` here?

core/jazz_t-vault/swagger/swagger.json
devsatishm: Please fix the description to reflect the purpose of the API. Based on the reply, it can be - "Get role details that's assigned to a safe"

core/jazz_t-vault/swagger/swagger.json
devsatishm: `Removes user from T-Vault` => `Remove user from a safe`?

SiniWilson: NO. Its an object. But in the format of "roles" : {"rolename": "permission"}. I couldn't find a proper way for map these key value pair in swagger though.

SiniWilson: NO. Its an object. But in the format of "users" : {"username": "permission"}. Same as below.

core/jazz_t-vault/components/utils/vault.js
SiniWilson: The T-vault response itself formatted, so there is no need for  further massage from our side. But applied the logic to convert "aws-role" key to roles key in response.

core/jazz_t-vault/index.js
SiniWilson: We are returning the responses in the format of { data: {}, input:{}} , since here  we are taking safename  in the path as input, returning that in the response object.

core/jazz_t-vault/swagger/swagger.json
SiniWilson: Done

devsatishm: @SiniWilson Little confused, how will this response look if a safe has two roles with read access to the secrets in that safe. I expect this to be an array, something like below - 
```
{
... 
"users": [{"username": "a", "permission": "read"}, {"username": "b", "permission": "write"}],
"roles": [{"rolename": "foo", "permission": "read"}, {"rolename": "bar", "permission": "read"}]
}
```

devsatishm: Same as roles question below. How will we represent multiple users with their permissions if it is not an array of user objects?

SiniWilson: Sample resp :   "users": {
            "sini": "write"
        },
        "roles": {
            "testrole": "write"
        }

devsatishm: Discussed with @SiniWilson, apparently, we will return this like below - 
```
{
... 
"users": {"userA": "read", "userB": "write"},
"users": {"roleA": "read", "roleB": "read"}
}```

PR #1160, opened 2019-09-24
Fixes to azure extension feature implementation

SauarvD:
### Requirements

* UI changes to name of certain fields
* Service on-boarding build pack changes, removal of extra EOF dot that was failing the jenkins


devsatishm:
> LGTM.
> 
> Only concern we are removing spaces in the Azure Region in the installer: [tmobile/jazz-installer#561](https://github.com/tmobile/jazz-installer/pull/561)

It looks like we're using `westus` instead of `West US` in the code which is what azure cli/api expects during deployment, which makes me feel that functionality should work as expected. Anyways, stack validation should catch this right away.

PR #1164, opened 2019-09-27
Updates to azure support

SauarvD:
### Description

* `azure_enabled` should be `false` when extension is not applied
* Disabled API option for Azure

### DIFF
--- a/builds/jenkins-build-pack-website/Jenkinsfile
+++ b/builds/jenkins-build-pack-website/Jenkinsfile
@@ -35,6 +35,7 @@ import groovy.transform.Field
 @Field def service_config
 @Field def websiteType
 @Field def azureAccount
+@Field def angularVersion
 
 node()  {
   try {
@@ -1163,6 +1164,7 @@ def validateAngularTemplate() {
 		def angularVersionString = new groovy.json.JsonSlurperClassic().parseText(packageString).dependencies.'@angular/core'
 
 		def version = angularVersionString.substring(1, angularVersionString.indexOf('.'))
+		angularVersion = version;
 		echo "Angular Project version $version"
 	} catch(ex) {
 		error "Error while checking version::"+ ex.getMessage();
@@ -1211,7 +1213,11 @@ def buildWebsite(framework) {
 				sh "rm -rf dist"
 				sh "npm cache clean --force"
 				sh "npm install"
-				sh "ng build --output-path=dist"
+				if(angularVersion >= 6){
+					sh "ng build --output-path=dist"
+				} else {
+					sh "ng build --prod --output-path=dist"
+				}
 			break
 			case "react":
 				sh "rm -rf build"
--- a/builds/jenkins-build-pack-website/Jenkinsfile
+++ b/builds/jenkins-build-pack-website/Jenkinsfile
@@ -35,7 +35,6 @@ import groovy.transform.Field
 @Field def service_config
 @Field def websiteType
 @Field def azureAccount
-@Field def angularVersion
 
 node()  {
   try {
@@ -201,7 +200,10 @@ node()  {
 			if(service_config['framework'] == 'angular' || service_config['framework'] == 'react') {
 				events.sendStartedEvent('BUILD', 'build starts', context_map, environment_logical_id)
 				try {
-					buildWebsite(service_config['framework'])
+					def packageJson = readFile("app/package.json")
+					def angularVersionData = new groovy.json.JsonSlurperClassic().parseText(packageJson).dependencies.'@angular/core'
+					def version = angularVersionData.substring(1, angularVersionData.indexOf('.'))
+					buildWebsite(service_config['framework'], version)
 				} catch(ex) {
 					events.sendFailureEvent('BUILD', ex.getMessage(), context_map, environment_logical_id)
         			events.sendFailureEvent('UPDATE_DEPLOYMENT', ex.getMessage(), environmentDeploymentMetadata.generateDeploymentMap("failed", environment_logical_id, gitCommitHash), environment_logical_id)
@@ -1164,7 +1166,7 @@ def validateAngularTemplate() {
 		def angularVersionString = new groovy.json.JsonSlurperClassic().parseText(packageString).dependencies.'@angular/core'
 
 		def version = angularVersionString.substring(1, angularVersionString.indexOf('.'))
-		angularVersion = version;
+		def angularVersion = version;
 		echo "Angular Project version $version"
 	} catch(ex) {
 		error "Error while checking version::"+ ex.getMessage();
@@ -1205,7 +1207,7 @@ def validateReactTemplate() {
 /**
   * Function to do build of angular/react
 */
-def buildWebsite(framework) {
+def buildWebsite(framework, angularVersion) {
 	def websiteCodeFolder = "app/"
 	dir(websiteCodeFolder) {
 		switch(framework) {
--- a/core/jazz_ui/src/app/pages/environment-detail/environment-detail.component.ts
+++ b/core/jazz_ui/src/app/pages/environment-detail/environment-detail.component.ts
@@ -231,11 +231,7 @@ export class EnvironmentDetailComponent implements OnInit {
               return asset.asset_type === 'swagger_url';
             });
             if (foundAsset) {
-              if(foundAsset.provider == 'azure'){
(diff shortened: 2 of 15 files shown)


core/jazz_ui/src/app/pages/environment-detail/environment-detail.component.ts
```diff
@@ -231,7 +231,11 @@ export class EnvironmentDetailComponent implements OnInit {
               return asset.asset_type === 'swagger_url';
             });
             if (foundAsset) {
-              return window.open(environment.urls['swagger_editor'] + foundAsset.provider_id);
+              if(foundAsset.provider == 'azure'){
```
raghits: why this difference?

devsatishm: Agreed, this experience should be consistent irrespective of the provider.

SauarvD: reverted back the changes

builds/jenkins-build-pack-website/Jenkinsfile
```diff
@@ -35,6 +35,7 @@ import groovy.transform.Field
 @Field def service_config
 @Field def websiteType
 @Field def azureAccount
+@Field def angularVersion
```
devsatishm: Defining this as a global var can affect builds on other user apps. Let's scope it to local by default.

builds/jenkins-build-pack-website/Jenkinsfile
```diff
@@ -1211,7 +1213,11 @@ def buildWebsite(framework) {
 				sh "rm -rf dist"
 				sh "npm cache clean --force"
 				sh "npm install"
-				sh "ng build --output-path=dist"
+				if(angularVersion >= 6){
+					sh "ng build --output-path=dist"
+				} else {
+					sh "ng build --prod --output-path=dist"
```
devsatishm: Why is this always `--prod`? Shouldn't' this be environment specific?

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.html
```diff
@@ -20,9 +20,9 @@ <h4 class="bold-title">
                     <p class="step-title">Choose the type of service you want to create<span>*</span></p>
                 </div>
                 <div class="space-between content">
-                    <section class="pad-left0 service-box-section">
-                        <div tabindex="0" class="service-box" [ngClass]="{'service-box-selected': typeOfService == 'api'}" (keyup.enter)="changeServiceType('api')" (click)="changeServiceType('api')">
-                            <div class="icon-icon-api selected"></div>
+                    <section class="pad-left0 service-box-section disabled-box">
```
devsatishm: This should be disabled only when `provider = azure`.

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.ts
```diff
@@ -396,7 +396,7 @@ export class CreateServiceComponent implements OnInit {
       this.typeOfPlatform = platformType;
       if(this.typeOfPlatform == 'azure'){
         if(this.typeOfService == 'sls-app'){
-          this.changeServiceType('api')
+          this.changeServiceType('function')
```
devsatishm: What's the purpose of this code? It is confusing since you are setting the service type as `function` when it is `sls-app`?

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.ts
```diff
@@ -839,7 +839,7 @@ export class CreateServiceComponent implements OnInit {
     this.submitted = true;
     this.getData();
     this.createService();
-    this.typeOfService = 'api';
+    this.typeOfService = 'function';
```
devsatishm: What's the purpose of this change?

builds/jenkins-build-pack-website/Jenkinsfile
SauarvD: the command with the --prod flag was already present in the develop branch when I was merging it with azure support branch. And the reason for two different commands is to support new websites created with angular templates which are upgraded to 7 and to also support older angular websites of version 2.

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.ts
SauarvD: This was done on purpose based on edge case. If user selects sls-app service type and then selects AWS, it's fine. But if he wishes to change the platform from AWS to AZURE, the serviceType is still sls-app, which is not supported, that's why under this case, the serviceType is default to API, but since we are disabling api so by default it is function.

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.ts
SauarvD: So by default the service type is API, whenever you open the create-service window. Since we are disabling api now, so it changed to function. The way it is changed is wrong. It has to be this way ONLY if azure is the provider.

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.html
SauarvD: Done

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.ts
SauarvD: Done

builds/jenkins-build-pack-website/Jenkinsfile
SauarvD: Done. Now passed as local variable

builds/jenkins-build-pack-website/Jenkinsfile
```diff
@@ -200,7 +200,10 @@ node()  {
 			if(service_config['framework'] == 'angular' || service_config['framework'] == 'react') {
 				events.sendStartedEvent('BUILD', 'build starts', context_map, environment_logical_id)
 				try {
-					buildWebsite(service_config['framework'])
+					def packageJson = readFile("app/package.json")
+					def angularVersionData = new groovy.json.JsonSlurperClassic().parseText(packageJson).dependencies.'@angular/core'
+					def version = angularVersionData.substring(1, angularVersionData.indexOf('.'))
```
raghits: What happens if the build agent i.e. global angular cli has a different version of angular cli? Isn't that what counts here as we are using `ng build` for the build which uses global angular cli?

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.ts
raghits: This needs to be cleaned up. If we do just top down filtering of options rather than going both ways.

core/jazz_ui/src/app/secondary-components/create-service/oss/create-service.component.ts
```diff
@@ -164,6 +164,7 @@ export class CreateServiceComponent implements OnInit {
   public apigeeFeature = this.buildEnvironment.INSTALLER_VARS.feature.apigee && this.buildEnvironment.INSTALLER_VARS.feature.apigee.toString() === "true" ? true : false;
   public selectedDeploymentTarget = "aws_apigateway";
   public azureEnabled: boolean = false;
+  public disableFunction: boolean = false;
```
raghits: what is the purpose of this variable?

SauarvD: since there was no parameter to detect whether functions with NONE are selected or not and now we are NOT creating functions with NONE, that's why this variable is being introduced to take care of that scenario.

builds/jenkins-build-pack-website/Jenkinsfile
SauarvD: yes but in order to support the older services with the same build pack, if else condition was introduced for building older and newer angular versions of websites.

raghits: Now, if the angular cli in the package.json is 5.x, the angular cli in global is 7.x, u will be running with a new version of angular in global along with the --prod flag which will fail. Right?

PR #1168, opened 2019-10-04
Address metrics not displaying

pgobin:


core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
```diff
@@ -536,19 +536,12 @@ export class ServiceMetricsComponent implements OnInit, AfterViewInit {
   setAsset() {
     switch (this.serviceType) {
       case 'api':
-        let method = this.filters.getFieldValueOfLabel('METHOD');
```
raghits: Is this not needed any more? So this switch statement is probably not needed?

PR #1174, opened 2019-10-09
metrics support changes to azure-support

SauarvD:
### Requirements

* Fixed website metrics which was not working.
* Made dynamic filters for metrics based on providers.

### DIFF
--- a/core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
+++ b/core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
@@ -89,6 +89,7 @@ export class ServiceMetricsComponent implements OnInit, AfterViewInit {
   errMessage: any;
   private toastmessage: any = '';
   private slsLambdaselected;
+  public provider: any;
   constructor(private request: RequestService,
     private utils: UtilsService,
     private messageservice: MessageService,
@@ -126,6 +127,9 @@ export class ServiceMetricsComponent implements OnInit, AfterViewInit {
   }
 
   ngOnInit() {
+    if(this.service.provider == 'azure'){
+      this.setFilters();
+    }
     this.serviceType = this.service.type || this.service.serviceType;
     this.setPeriodFilters();
   }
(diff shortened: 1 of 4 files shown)


core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
```diff
@@ -160,6 +164,13 @@ export class ServiceMetricsComponent implements OnInit, AfterViewInit {
     }
   }
 
+  setFilters(){
+    // we are targeting the second object in the FormFields where values for AGGREGATION is there
+    this.formFields[2].options = ['Average', 'Total', 'Maximum']
```
raghits: Is this accurate? We are seeing mixed case here. Do we need to set both value and values?

core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
```diff
@@ -447,7 +459,9 @@ export class ServiceMetricsComponent implements OnInit, AfterViewInit {
         break;
       case 'website':
         let websiteAssets = _(this.queryDataRaw.assets).map('type').uniq().value();
-        this.filters.addField('Filter By:', 'ASSET', websiteAssets, 'select', null, 'cloudfront');
+        if (this.provider != 'azure'){
```
raghits: Will this work if we add another provider. Should we just say `this.provider == 'aws'` instead?

core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
SauarvD: value and values both are necessary as the filter modal component being used there needs both of these fields

core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
SauarvD: My Bad. Changed to provider == 'aws'

PR #1178, opened 2019-10-11
Updates to documentation related to azure support

devsatishm:
## Description

Updates to documentation related to new features


devsatishm:
> LGTM. But are mentioning anywhere that API for azure we are not supporting as for now. Will be releasing that later.

This should be fine, changes will be going in as part of a patch.

PR #1180, opened 2019-10-11
 Fixing logs issue related to elastic search version upgrade

SiniWilson:
### Requirements

 Fixing logs issue related to elastic search version upgrade

### Description of the Change

Changed mapping pattern in es-kinesis-logs-streamer
Changed request format in logs api

### DIFF
--- a/core/jazz_es-kinesis-log-streamer/index.js
+++ b/core/jazz_es-kinesis-log-streamer/index.js
@@ -14,7 +14,7 @@
 // limitations under the License.
 // =========================================================================
 
-var https = require('https');
+var http = require('http');
 var zlib = require('zlib');
 
 const errorHandlerModule = require("./components/error-handler.js");
@@ -98,8 +98,8 @@ function handler(event, context, cb) {
 };
 
 function post(config, body, callback) {
-  let requestParams = utils.buildRequest(config.ES_ENDPOINT, body);
-  let request = https.request(requestParams, function (response) {
+  let requestParams = utils.buildRequest(config, body);
+  let request = http.request(requestParams, function (response) {
 
     let responseBody = '';
     response.on('data', function (chunk) {
--- a/core/jazz_es-kinesis-log-streamer/index.js
+++ b/core/jazz_es-kinesis-log-streamer/index.js
@@ -14,8 +14,8 @@
 // limitations under the License.
 // =========================================================================
 
-var http = require('http');
 var zlib = require('zlib');
+var request = require('request');
 
 const errorHandlerModule = require("./components/error-handler.js");
 const responseObj = require("./components/response.js");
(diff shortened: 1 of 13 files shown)


core/jazz_es-kinesis-log-streamer/index.js
```diff
@@ -14,7 +14,7 @@
 // limitations under the License.
 // =========================================================================
 
-var https = require('https');
+var http = require('http');
```
devsatishm: Can we use [request](https://github.com/request/request) module instead of this http <-> https switch? `request` module should take care of this for us. Covers use case when jazz admin replaces es_endpoint with https or http endpoint value..

core/jazz_logs/index.js
```diff
@@ -133,9 +134,10 @@ module.exports.handler = (event, context, cb) => {
 
       var req = utils.requestLoad;
       req.url = config.KIBANA_URL + "/elasticsearch/_msearch";
-      req.body = setRequestBody(servCategory, env, querys, startTime, endTime, size, page);
+      req.body = setRequestBody(servCategory, querys, filterQuerys, startTime, endTime, size, page);
 
       request(req, function (err, res, body) {
+        logger.info("RESPONSE FROM ES : " + JSON.stringify(res));
```
devsatishm: logger.debug

core/jazz_logs/config/test-config.json
```diff
@@ -2,53 +2,79 @@
...
+  "LOG_LEVELS": [
+    {
       "Level": 0,
-      "Type": "off"
+      "Type": "OFF"
     },
     {
       "Level": 1,
-      "Type": "fatal"
+      "Type": "FATAL"
```
devsatishm: Why this switch?

core/jazz_logs/index.js
```diff
@@ -59,7 +59,7 @@ module.exports.handler = (event, context, cb) => {
         return cb(JSON.stringify(errorHandler.throwInputValidationError("missing required input parameter category.")));
       }
 
-      if (!event.body.type || !_.includes(config.VALID_LOGTYPES, event.body.type.toLowerCase())) {
```
devsatishm: I would prefer keeping this as-is. This would handle case insensitive searches (which is nice!).

core/jazz_logs/index.js
```diff
@@ -183,37 +185,38 @@ module.exports.handler = (event, context, cb) => {
...
 
+
     var params = {
       "size": size,
       "from": page,
       "sort": [{
         "timestamp": {
-          "order": "desc"
+          "order": "desc",
+          "unmapped_type": "boolean"
```
devsatishm: What does this do?

core/jazz_logs/config/test-config.json
raghits: Everything seems to have changed from lowercase to uppercase?

core/jazz_logs/index.js
raghits: Right. In which case there is no need to change the case of the VALID_LOGTYPES.

core/jazz_logs/index.js
raghits: https://www.elastic.co/guide/en/elasticsearch/reference/6.3/search-request-sort.html#_ignoring_unmapped_fields

raghits: @SiniWilson what happens if unmapped_type is not set or is set to a different data type such as `date` or `long`?

devsatishm: This is not needed at all since timestamp is already mapped to type - `date` for this index. 

See https://github.com/tmobile/jazz-installer/blob/master/installer/terraform/scripts/configureESEndpoint.sh#L20 &  https://github.com/tmobile/jazz/blob/master/core/jazz_es-kinesis-log-streamer/_ES/apilogs.json

PR #1181, opened 2019-10-11
Graphs, filters and carousel's UI is updated with bug fixes.

dmpl1:
### Requirements

* Implement New Chart Across Application

### Description of the Change

* Graphs update using Canvas.
* Filter Issues fixes
* Carousel CSS changes.

### DIFF
--- a/core/jazz_ui/src/app/pages/environment-logs/env-logs-section.component.ts
+++ b/core/jazz_ui/src/app/pages/environment-logs/env-logs-section.component.ts
@@ -545,7 +545,11 @@ export class EnvLogsSectionComponent implements OnInit {
 		if(response&&response.data&&response.data.assets){
 			this.assetsNameArray.push(response);
 			let assets=_(response.data.assets).map('asset_type').uniq().value();
-
+			const filterWhitelist = [	
+				'lambda',
+				'apigateway'	
+			];
+			assets = assets.filter(item => filterWhitelist.includes(item));
 			 let validAssetList = assets.filter(asset => (env_oss.assetTypeList.indexOf(asset) > -1));
 			 validAssetList.splice(0, 0, 'all');
 			this.responseArray = this.assetsNameArray[0].data.assets.filter(asset=>(validAssetList.indexOf(asset.asset_type)>-1));
--- a/core/jazz_ui/src/app/pages/environment-logs/env-logs-section.component.ts
+++ b/core/jazz_ui/src/app/pages/environment-logs/env-logs-section.component.ts
@@ -546,12 +546,6 @@ export class EnvLogsSectionComponent implements OnInit {
 			this.assetsNameArray.push(response);
 			let assets=_(response.data.assets).map('asset_type').uniq().value();
 
-			// TODO: Consider hoisting to member or configuration
-			const filterWhitelist = [
-				'lambda',
-			];
-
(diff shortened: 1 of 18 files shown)


core/jazz_ui/src/app/pages/environment-logs/env-logs-section.component.ts
```diff
@@ -546,12 +546,6 @@ export class EnvLogsSectionComponent implements OnInit {
 			this.assetsNameArray.push(response);
 			let assets=_(response.data.assets).map('asset_type').uniq().value();
 
-			// TODO: Consider hoisting to member or configuration
```
pgobin: Can you clarify why this is removed?

raghits: Without this, all asset types will show in the Asset Type filter for logs (similarly we have for metrics). Only some asset types have logs and metrics.

dmpl1: Though I have removed the above filter, there is a list of supported asset types provided in env.oss file, which is listed below. So asset names filter will be shown for all those asset types. If below all Asset types don't support for asset name filter, then I can re-add only for those we require it. 
assetTypeList: [
    "lambda",
    "apigateway",
    "cloudfront",
    "s3",
    "dynamodb",
    "cloudwatch_event",
    "dynamodb_stream",
    "iam_role",
    "sqs",
    "kinesis_stream",
    "apigee_proxy"
  ]

core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
```diff
@@ -585,21 +601,20 @@ export class ServiceMetricsComponent implements OnInit, AfterViewInit {
...
         };
-
-        if (!obj['y']) {
-          obj['y'] = parseInt(dataPoint[valueProperty.toLowerCase()])
-        }
         return obj;
       });
 
     let timeRange = this.filters.getFieldValueOfLabel('TIME RANGE');
+    this.aggregation = this.filters.getFieldValueOfLabel('AGGREGATION') == 'sum' ? 'Sum': 'Avg.';
```
raghits: Are we assuming Sum and Average are the only valid values? Shouldn't it be more generic or have a mapping instead?

dmpl1: Since in dropdown of aggregation, there are two values - sum and avg, so there are the only valid ones as per my understanding.

core/jazz_ui/src/app/pages/environment-logs/env-logs-section.component.ts
dmpl1: After discussion with Prakash, removed whitelist for asset types is re-added.

core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
raghits: @dmpl1 Once we bring in Azure, we have some others as well. Might be better to move this to a mapping.

raghits: Lets defer this to once Azure changes comes in.

PR #1182, opened 2019-10-11
:bug: Use correct service type for serverless apps

devsatishm:
### Description of the Change

Use correct service type for serverless apps in metrics UI component


core/jazz_ui/src/app/pages/service-metrics/service-metrics.component.ts
```diff
@@ -200,7 +200,7 @@ export class ServiceMetricsComponent implements OnInit, AfterViewInit {
               for (var i = 0; i < self.assetWithDefaultValue.length; i++) {
                 self.assetList[i] = self.assetWithDefaultValue[i].replace(/_/g, " ");
               }
-              if(self.service.serviceType == "sls-app"){
+              if(self.service.serviceType == "custom"){
```
raghits: Is `self.service.serviceType` coming from UI? `sls-app` needs to be translated to `custom` only finally where it surfaced in the UI. Similarly, any `custom` selection should be first translated to `sls-app`. The entire backend should still expect it to be `sls-app`.

devsatishm: @SauarvD FYI

SauarvD: @raghits @devsatishm in the backend and filters we are still using sls-app. This change came in because for the UI, we changed serviceType from sls-app to custom which got merged in the previous PR . And in metrics component, depending on the service type we are making decisions so since we changed sls-app -> custom, it was never going inside the required function.

raghits: ok.

PR #1184, opened 2019-10-14
changes for build-pack-website

SauarvD:
### Requirements

* version variable is applicable only for angular website. Added a condition for the same.

### DIFF
--- a/builds/jenkins-build-pack-website/Jenkinsfile
+++ b/builds/jenkins-build-pack-website/Jenkinsfile
@@ -201,9 +201,13 @@ node()  {
 				events.sendStartedEvent('BUILD', 'build starts', context_map, environment_logical_id)
 				try {
 					def packageJson = readFile("app/package.json")
-					def angularVersionData = new groovy.json.JsonSlurperClassic().parseText(packageJson).dependencies.'@angular/core'
-					def version = angularVersionData.substring(1, angularVersionData.indexOf('.'))
-					buildWebsite(service_config['framework'], version)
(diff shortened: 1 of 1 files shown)


builds/jenkins-build-pack-website/Jenkinsfile
```diff
@@ -201,9 +201,13 @@ node()  {
...
 					def packageJson = readFile("app/package.json")
-					def angularVersionData = new groovy.json.JsonSlurperClassic().parseText(packageJson).dependencies.'@angular/core'
-					def version = angularVersionData.substring(1, angularVersionData.indexOf('.'))
-					buildWebsite(service_config['framework'], version)
+					if(service_config['framework'] == 'angular'){
+						def angularVersionData = new groovy.json.JsonSlurperClassic().parseText(packageJson).dependencies.'@angular/core'
+						def version = angularVersionData.substring(1, angularVersionData.indexOf('.'))
+						buildWebsite(service_config['framework'], version)
+					} else {
+						buildWebsite(service_config['framework'], null)
```
raghits: This address the scenario for HTML5 and ReactJS?

SauarvD: for HTML5 (static template), it will never come inside this block as this block is only for angular and ReactJS. And we need the angular version for angular template only and for react we are not dependent on any version as such.

raghits: ok

PR #1196, opened 2019-10-17
Renaming git repo

SiniWilson:
### Requirements

Renaming git repo to become consistent

### Description of the Change

Renamed build-pack-lambda to build-pack-function


raghits:
@SiniWilson Do we need to change any reference to this?

PR #1198, opened 2019-10-17
css fix for button position.

dmpl1:
### Requirements

* sls-app issue fix

### Description of the Change

Test function button position fixed at service and environmental level

### DIFF
--- a/core/jazz_ui/src/app/pages/service-detail/service-detail.component.html
+++ b/core/jazz_ui/src/app/pages/service-detail/service-detail.component.html
@@ -19,7 +19,7 @@
                 <span *ngIf="service.runtime"> - {{service.runtime}}</span>
             </span>
             </div>
-            <div class="image-btn-div">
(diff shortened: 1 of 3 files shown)


core/jazz_ui/src/app/pages/service-detail/service-detail.component.html
```diff
@@ -19,7 +19,7 @@ <h1 class="page-hdr bold-title">{{service.name}}</h1>
                 <span *ngIf="service.runtime"> - {{service.runtime}}</span>
             </span>
             </div>
-            <div class="image-btn-div">
+            <div class="image-btn-div flex-end">
```
raghits: Why is this changed needed in the service detail? What issue is this addressing?

dmpl1: CSS class was added in service-detail as a precaution and to have code in sync between env-detail and service-detail.

PR #1201, opened 2019-10-18
Fixing logs issue related to ES version upgrade

SiniWilson:
### Requirements

Fixing logs issue related to ES version upgrade

### Description of the Change

Added asset_identifier
Added unmapped_type


core/jazz_logs/utils.js
```diff
@@ -50,22 +50,16 @@ var set_query = function (type, value) {
...
 		"query_string": {
-			"query": type + ":(" + value
+			"query": type + ":(" + value.toUpperCase()
 		}
 	};
 	var requestedLogType = LOG_LEVEL_CONFIG.filter(configObject => configObject.Type === value);
 	if (requestedLogType[0]) {
 		LOG_LEVEL_CONFIG.map(function (configObject) {
-			if (configObject.Level <= parseInt(requestedLogType[0].Level)) {
+			if (configObject.Level < parseInt(requestedLogType[0].Level)) {
```
raghits: Shouldn't this be `configObject.Level <= parseInt(requestedLogType[0].Level` like it was before?

raghits: Applied this to stack and validated. Works as expected. Closing this.

SiniWilson: configObject.Level <= parseInt(requestedLogType[0].Level will add current log level twice since we are adding current log level in 
https://github.com/tmobile/jazz/pull/1201/files/69273e3adb05fe44d9c984f5d944acbff17e0896#diff-ad216a653c09ea10d99831b7d8712aceR56. So we don't want the equal to sign

PR #1212, opened 2019-10-31
Adding validations and creating role in vault

SiniWilson:
### Requirements

Adding validations
Creating role in vault
Massaging get safe response


core/jazz_t-vault/components/utils/validations.js
```diff
@@ -91,6 +91,24 @@ function validateUserInVaultDeleteInput(event) {
...
 
+function validateRoleArn(arn) {
+  return new Promise((resolve, reject) => {
+    let regEx = new RegExp("^arn:aws:iam::\\d{12}:role/?[a-zA-Z_0-9+=,.@\\-_/]+");
+    if (regEx.test(arn)) return resolve();
+    else return reject({ "errorType": "inputError", "message": "The provided arn is not valid - " + arn });
+  });
+}
+
+function validateFieldLength(data) {
```
raghits: Make this generic so it can be used to check for fields of different minimum lengths.
function validateFieldLength(data, minLength)

eg. name is 3 characters
description is 10 characters

raghits: Also pass just one field instead of all the fields i.e. not data but data.name , data.description separately.

core/jazz_t-vault/components/utils/vault.js
```diff
@@ -401,7 +458,9 @@ module.exports = {
   createUserInSafe,
   deleteUserFromSafe,
   getRoleInSafe,
+  createRole,
```
raghits: createRole combines createRoleInVault and createRoleInSafe?
Also, does it check to see if role already exists in vault and safe before creating?

similarly we need createUser which combines createUserInVault and createUserInSafe.
Again it should check to see if user already exists in vault and safe before creating it.

core/jazz_t-vault/components/utils/validations.js
SiniWilson: Done

core/jazz_t-vault/components/utils/vault.js
SiniWilson: Done

raghits: This will be addressed in a separate story.

raghits:
@SiniWilson Is this still WIP? Also, is this applied to current stack?

PR #1214, opened 2019-10-31
Fix for GET  /services & added new API path - /services/search

dmpl1:
New endpoint for search and provided exact match for the services list.

### Requirements
* Defect: GET  /services retrieves multiple services during search since it uses CONTAINS instead of exact match in DynamoDB query

### Description of the Change
* Created a new endpoint for search. 
* Provided exact match for services list.


core/jazz_services/swagger/swagger.json
```diff
@@ -542,6 +542,537 @@
...
+              "responseParameters": {
+                "method.response.header.Access-Control-Allow-Headers": "'Content-Type,X-Amz-Date,Authorization,X-Api-Key'",
+                "method.response.header.Access-Control-Allow-Methods": "'*'",
+                "method.response.header.Access-Control-Allow-Origin": "'*'"
+              }
+            }
+          }
+        }
+      },
+      "post": {
```
devsatishm: Why do we need `post`?

devsatishm:
Looks good overall. Please fix the Travis build error (unit tests).

core/jazz_services/swagger/swagger.json
dmpl1: Post is not needed. Changes updated.

core/jazz_services/components/crud/getSearchList.js
```diff
@@ -0,0 +1,159 @@
+// =========================================================================
```
raghits: getSearchList.js is practically the same as getList.js except probably for one line - reuse the code and make it DRY?

core/jazz_services/swagger/swagger.json
```diff
@@ -542,6 +542,332 @@
         }
       }
     },
+    "/services/search": {
+      "options": {
+        "description": "CORS Support",
```
raghits: Had to enable CORS manually in API Gateway for /services/search.

core/jazz_services/swagger/swagger.json
```diff
@@ -542,6 +542,332 @@
...
+        "consumes": [
```
raghits: See POST method for /services/search which is not needed. 
image

dmpl1: Removed POST

core/jazz_services/components/crud/getSearchList.js
dmpl1: Addressed with conditions.

PR #1223, opened 2019-11-14
Added missing path check for /search.

dmpl1:
### Requirements

* GET search fix for exact match and a new path for search created for service list.

### Description of the Change

* Added missing path check for /search before adding a condition for filter match type ( i.e., contains search/ exact search) to fetch results.

### DIFF
--- a/core/jazz_services/index.js
+++ b/core/jazz_services/index.js
@@ -125,7 +125,9 @@ module.exports.handler = (event, context, cb) => {
                     // fetch services list from dynamodb, filter if required
                     fetchServices: function (onComplete) {
                         var query = event.query;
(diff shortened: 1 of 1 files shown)


raghits:
Validation issue missing pagination. Please address.

dmpl1:
> Can you check if the webapp services list is using this path: `/services/search`?

dmpl1:
@raghits @devsatishm  stack updated with changes, validated and webapp services list is also using 'services/search' path.  
Validation issue of missing pagination is observed in secret-mgmt branch too. so looks like this PR has no relation with pagination and can be addressed separately.

PR #1225, opened 2019-11-18
Secret management support in ACL API

SiniWilson:
### Requirements

Secret management support in ACL API

### Description of the Change

When user is added/removed to 'admin' group/role, give/remove access to all the safes for a service.


core/jazz_acl/index.js
```diff
@@ -69,6 +73,12 @@ async function processACLRequest(event, config) {
       }
 
       await exportable.processScmPermissions(config, serviceId, policies, 'add');
+
+      logger.info("vault is enabled. global.globalConfig.VAULT.IS_ENALED : " + global.globalConfig.VAULT.IS_ENABLED)
```
raghits: nitpick: typo 'IS_ENABLED`

core/jazz_acl/index.js
```diff
@@ -69,6 +73,12 @@ async function processACLRequest(event, config) {
...
+      if (global.globalConfig.VAULT.IS_ENABLED) {
```
raghits: Do we need a check if `global.globalConfig.VAULT` is null? Make sure this doesn't fail if T-Vault extension is not installed.

core/jazz_acl/components/utils/vault.js
```diff
@@ -0,0 +1,206 @@
...
+    let existingSafeUsers = [];
+    let processEachAddOrRemoveUser = [];
+    if (safeDetails && safeDetails.users) {
+      for (let user in safeDetails.users) {
+        existingSafeUsers.push(user);
+      }
+    }
+
+    logger.info("existingSafeUsers " + existingSafeUsers)
+    let usersTobeAdded = adminUsers.filter(adminUser => !existingSafeUsers.includes(adminUser));
```
raghits: nitpick. `usersToBeAdded`, `usersToBeRemoved`

raghits:
@SiniWilson Is this still WIP? Also, is this validated? If so, which stack can this be validated in?

core/jazz_acl/index.js
SiniWilson: We don't need the check as it is coming from the config file. Also  it will have a false value if vault is not enabled

core/jazz_acl/index.js
SiniWilson: Removed the log since we don't want the log over here

core/jazz_acl/components/utils/vault.js
SiniWilson: Done

core/jazz_acl/config/global-config.json
```diff
@@ -1,4 +1,27 @@
 {
+  "CASBIN": {
```
devsatishm: These can be environment specific. They should be living in $env-config.json.

SiniWilson: Yes. But I saw we are configuring one DB and loading it for all the environments. That's why I changed that. Please see the configuration
https://github.com/tmobile/jazz/blob/master/builds/jazz-build-module/service-configuration-data-loader.groovy#L62

SiniWilson: Changed CASBIN as environment specific

PR #1211, opened 2019-10-30
Feature/tvault safe

dmpl1:
### Requirements
 * create safe, add user to safe and delete safe APIs integration.

### Description of the Change

* Added create safe related functionality at inital-commit and create-branch level.
* added delete safe functionality at delete-branch level
* stored safe-details at metadata.

PR #1227, opened 2019-11-21
Create/Delete dedicated role as part of service deployment/deletion

SiniWilson:
### Requirements

 Create/Delete dedicated role as part of service deployment/deletion

### Description of the Change

- Adding custom role to safe 
- Should only work if T-Vault feature extension is enabled


builds/jenkins-build-pack-function/Jenkinsfile
```diff
@@ -1268,8 +1307,8 @@ def isExistingLogGroup (groupName, credsId){
 	def filterLogsJSON = toJSON.parseText(filterLogs)
 	def logGroups = filterLogsJSON.logGroups
 	if(logGroups.size() > 0  && logGroups[0].logGroupName == groupName){
-		isExistingLogGrp = true
-	}
+		isExistingLogGrp = trues
```
raghits: typo: `trues`?

SiniWilson: My bad. I think the save- ctrl+s caused the issue. Fixed

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -41,430 +42,460 @@ import groovy.transform.Field
...
-
-  }//dir ends here
+        events.sendCompletedEvent('UPDATE_ENVIRONMENT', 'Environment update event for deployment completion', environmentDeploymentMetadata.generateEnvironmentMap("deployment_completed", environment_logical_id, env_metadata), environment_logical_id)
+        events.sendCompletedEvent('UPDATE_DEPLOYMENT', "Deployment completion event for ${env_key} deployment", environmentDeploymentMetadata.generateDeploymentMap("successful", environment_logical_id, gitCommitHash), environment_logical_id)
+        events.sendCompletedEvent('DEPLOY_TO_AWS', 'Successfully deployed services to AWS', serviceContext, environment_logical_id)
+      }//stage of deployment to an environment ends here
+
+      //Udating safe details
+        if (config['domain'] != 'jazz' && configLoader.TVAULT && configLoader.TVAULT.IS_ENABLED) {
+          stage('Update Safe') {
```
devsatishm: Now, we will see this stage always. Shouldn't we show this stage only when the current role is different from role in env/safe metadata?

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -41,430 +42,460 @@ import groovy.transform.Field
...
+            events.sendStartedEvent('UPDATE_SAFE', 'safe updation started', context_map, environment_logical_id)
+            withCredentials([
```
devsatishm: Why do we need this block? Are we using aws cli in this block?

builds/jazz-build-module/environment-deployment-metadata-loader.groovy
```diff
@@ -38,13 +39,14 @@ def initialize(serviceConfig, configLoader, scmMdule, branch, buildUrl, buildId,
 }
 
 def getEnvironmentLogicalId() {
-	if (g_environment_logical_id == null && g_service_config['domain'] != "jazz") {
+	if ((g_environment_logical_id == null || g_environment_details == null) 
+								&& g_service_config['domain'] != "jazz") {
```
devsatishm: Why do we need to filter based on domain? This should work for all services irrespective of the domain.

builds/jazz-build-module/environment-deployment-metadata-loader.groovy
```diff
@@ -70,15 +73,14 @@ def getEnvironmentLogicalId() {
 	return g_environment_logical_id
 }
 
-def getEnvironmentInfo() {
+def getEnvironmentInfo(g_environment_logical_id) {
```
devsatishm: `g_environment_logical_id` is a global var which is assumed to be available already (see `getEnvDeploymentDescriptor`). Do we need to pass this var?

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -41,430 +42,460 @@ import groovy.transform.Field
...
+              try {
+                // initialize aws credentials
+                def randomString = utilModule.generateRequestId();
+                credsId = "jazz-${randomString}";
+                sh "aws configure set profile.${credsId}.region ${service_config.region}"
+                sh "aws configure set profile.${credsId}.aws_access_key_id $AWS_ACCESS_KEY_ID"
+                sh "aws configure set profile.${credsId}.aws_secret_access_key $AWS_SECRET_ACCESS_KEY"
+            
+                def environment = environmentDeploymentMetadata.getEnvironmentInfo(environment_logical_id)
+                if (environment && environment.metadata && environment.metadata.safe_details) {
```
devsatishm: Suggestion: `safe_details` => `safe`.

builds/jazz-build-module/vault-utility-module.groovy
```diff
@@ -0,0 +1,114 @@
...
+	def safeDetails
+	 try {
+		safeDetails = parseJson(safeResponce)
+	} catch (ex) {
+		echo "Error in getting safe details. ${ex}"
+	}
+	return safeDetails
+}
+
+def getRoleDetails(lambdaARN, credsId) {
```
devsatishm: It would be better if we just persist the role that we create for the environment in its metadata instead of making one more call to AWS lambda service.

builds/jenkins-build-pack-function/Jenkinsfile
```diff
@@ -37,376 +37,410 @@ import groovy.transform.Field
...
+      } //end of deployment to an environment
 
-          } catch (ex) {
-            handleFail(ex, context_map, environment_logical_id, config, gitCommitHash)
-          } finally {
-            // reset Credentials
-            resetCredentials(credsId)
+      //Udating safe details
+      if (config['domain'] != 'jazz' && configLoader.TVAULT && configLoader.TVAULT.IS_ENABLED) {
+        stage('Update Safe') {
```
devsatishm: We will need to add roles to the safe before the deployment. Imagine RoleA is being used for a function. User supplied RoleB to be used for the function. If we deploy the service with RoleB and fail to add RoleB to the safe, function will now fail to access the secrets that it could access previously.

builds/jazz-build-module/vault-utility-module.groovy
devsatishm: If user supplies custom role, we already have a handle of it during the deployment. We're fetching list of roles having access to the safe previously. Compare the user supplied role with roles attached to the safe and if we don't see it, just add it to the safe.

builds/jenkins-build-pack-api/Jenkinsfile
devsatishm: Can be removed if this approach works - https://github.com/tmobile/jazz/pull/1227/files?utf8=%E2%9C%93&diff=unified&w=1#r351046001

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -41,430 +42,460 @@ import groovy.transform.Field
...
+                  def safeName = environment.metadata.safe_details.name
+                  def updatedIamRoleARN = vaultUtilityModule.updateSafeDetails(safeName, lambdaARN, credsId)
+                  events.sendCompletedEvent('CREATE_ASSET', null, utilModule.generateAssetMap(config['provider'], updatedIamRoleARN, "iam_role", config), environment_logical_id);
```
devsatishm: We will need to send CREATE_ASSET only when the platform creates the role. If user supplied custom role, we don't track it as an asset that the platform creates.

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -830,41 +861,45 @@ def updateOperationId(swaggerFileContent) {
 def writeServerlessFile(config, environment_logical_id, accountDetails){
 
   def iamRoleArnValue;
-  if(config['domain'] == 'jazz'){
-    iamRoleArnValue = config['iamRoleARN']
-  } else {
-    iamRoleArnValue = accountDetails.IAM.USERSERVICES_ROLEID;
-  }
+  if(config['domain'] == 'jazz') iamRoleArnValue = config['iamRoleARN']
```
devsatishm: This has to be updated a little bit. 
- `accountDetails.IAM.USERSERVICES_ROLEID` will never be used in the workflows again. We either create a dedicated role per user service/env & used it OR just use user provided role.
- Why is `iamRoleArnValue = config['iamRoleARN']` only  when `domain == jazz`? It should be true for all services from now on. If user provides custom role, we have it use it during safe updates and deployments

builds/serverless-config-pack/serverless-go.yml
```diff
@@ -102,6 +102,12 @@ functions:
 #Start:resources
 resourcesDisabled:
   Resources:
+    #Start:isVaultEnabled
+    customEventRole: ${file(policyFile.yml):vaultRole}
```
devsatishm: `customEventRole` is definitely misleading.. :)

devsatishm: Shouldn't this just be `basicRole` &  `vpcRole` instead?

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -830,41 +861,45 @@ def updateOperationId(swaggerFileContent) {
...
     sh "sed -i -- 's/\${file(deployment-env.yml):mainClass}/${config['mainClass']}/g' serverless.yml"
   }
-}
 
-/**
-	Set the Lambda execution role. If user provided a custom role override the default.
-*/
-def setLambdaExecutionRole(role) {
-  sh "sed -i -- 's#DEFAULT_LAMBDA_EXE_ROLE#${role}#g' serverless.yml"
+  if (config['domain'] != 'jazz' && configLoader.TVAULT && configLoader.TVAULT.IS_ENABLED ) {
```
devsatishm: I think this can be simplified further. As a best practice, we can create dedicated roles per service/environment irrespective of whether vault feature is enabled or not. 

If vault is enabled, add role to the safe
If vault is disable, do nothing.

cc @raghits

builds/jenkins-build-pack-api/Jenkinsfile
devsatishm: See my comments https://github.com/tmobile/jazz/pull/1227/files?utf8=%E2%9C%93&diff=unified&w=1#r351057997 on why this works if vault is enabled/disabled.

builds/jazz-build-module/environment-deployment-metadata-loader.groovy
SiniWilson: "g_environment_logical_id"  is not initialized in the case of master branch deployment. Also the "physical_id" is same for both stg and prod environments, we need to pass the logical id to get the exact environment details.  "g_environment_logical_id" I will change it to a local variable.

builds/jazz-build-module/vault-utility-module.groovy
SiniWilson: I am getting the role details from the lambda to keep it consistent for api, function and sls-app. In the case of sls-app role arn is available only after the deployment. Also there is chance for different roles for each lambda in sls-app. Also in sls-app the role names are generated by the serverless if the user not specified.

SiniWilson: Since we can have multiple role arns per service,  should we persist them in catalog ? In order to track the jazz created roles we can use the asset table.

builds/jenkins-build-pack-api/Jenkinsfile
SiniWilson: We still need the call to aws lambda see explanation above.

builds/jenkins-build-pack-api/Jenkinsfile
SiniWilson: In oss platform services(jazz core) have elevated permissions hence iamRoleArnValue = config['iamRoleARN'] only when domain == jazz

builds/jenkins-build-pack-api/Jenkinsfile
SiniWilson: We can make this change after merging all the open PRs for secret management.

builds/jenkins-build-pack-function/Jenkinsfile
SiniWilson: Please see the explanation above

builds/jazz-build-module/environment-deployment-metadata-loader.groovy
SiniWilson: Initially it used. Now not using. So removed

builds/jazz-build-module/environment-deployment-metadata-loader.groovy
SiniWilson: Changed to local variable

builds/serverless-config-pack/serverless-go.yml
SiniWilson: Changed

builds/jazz-build-module/environment-deployment-metadata-loader.groovy
SiniWilson: Had found another work around for this.  Please see the latest change.

builds/jenkins-build-pack-api/Jenkinsfile
SiniWilson: Addressed

builds/jenkins-build-pack-api/Jenkinsfile
raghits: @SiniWilson @devsatishm This change is pending support for dedicated roles per environment.

core/jazz_acl/components/utils/vault.js
```diff
@@ -185,8 +185,8 @@ function addOrRemoveAllAdminUsersToSafe(environmentDetails, configData, serviceI
 
     for (let key in environments) {
       const environment = environments[key];
-      if (environment.metadata && environment.metadata.SAFE_DETAILS)  //TODO to check the case after the environment changes added
-        processEachPromises.push(processEachPromise(environment.metadata.SAFE_DETAILS.name, configData, serviceId, authToken, adminUsers));
+      if (environment.metadata && environment.metadata.SAFE)  //TODO to check the case after the environment changes added
```
raghits: should this be `environment.metadata.safe`?

core/jazz_acl/components/utils/vault.js
```diff
@@ -185,8 +185,8 @@ function addOrRemoveAllAdminUsersToSafe(environmentDetails, configData, serviceI
...
+        processEachPromises.push(processEachPromise(environment.metadata.SAFE.name, configData, serviceId, authToken, adminUsers));
```
raghits: should this be `environment.metadata.safe`?

core/jazz_acl/components/utils/vault.js
SiniWilson: While validating with "safe" it was throwing error. So given as "SAFE" and at that time the other PR https://github.com/tmobile/jazz/pull/1211 not developed. So added TODO for  checking the case. Waiting for the Pr to be merged and I will add the case appropriately.

core/jazz_acl/components/utils/vault.js
SiniWilson: same as above explanation

PR #1228, opened 2019-11-25
Hotfix: Setting Default ProviderTimeout for api and function

raghits:
### Requirements

* Provide providerTimeout: 160 default for api and function

### Description of the Change

* Sets the default providerTimeout to 160 seconds for API and Function

### Benefits

Doesn't break the build

### Possible Drawbacks

None

### Applicable Issues

None


core/jazz_create-serverless-service/index.js
```diff
@@ -351,6 +351,19 @@ var getServiceData = (service_creation_data, authToken, configData, deploymentTa
     if (service_creation_data.service_type === "api" || service_creation_data.service_type === "function" || service_creation_data.service_type === "sls-app") {
       serviceMetadataObj.providerRuntime = service_creation_data.runtime;
     }
+    
+    //Adding providerTimeout key in service catalog
+    if (service_creation_data.service_type === "api" || service_creation_data.service_type === "function") {
+      if (service_creation_data.timeout)
```
devsatishm: User won't provide the timeout value during service creation. We can skip this check.

core/jazz_create-serverless-service/index.js
```diff
@@ -351,6 +351,19 @@ var getServiceData = (service_creation_data, authToken, configData, deploymentTa
...
+      {
+        serviceMetadataObj.providerTimeout = service_creation_data.timeout;
+      }
+      else
+      {
+        // default to 160 seconds
+        serviceMetadataObj.providerTimeout = 160;
```
devsatishm: Move this to configuration. Also, 30 sec is a good start (API Gateway max time out is 29 sec).

raghits: Updated.

core/jazz_create-serverless-service/index.js
raghits: Updated.

devsatishm:
Changes are tested in a trial environment, looks good!

PR #1248, opened 2019-12-13
Adding user policies while environment creation

SiniWilson:
### Requirements

Adding user policies while environment creation

### Description of the Change

Added policies updation in handler


builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -591,74 +590,73 @@ def checkoutSCM(repo_name) {
  * @param  serviceName
  * @return
  */
-def updateServiceNameConfig() {
```
raghits: `updateServiceNameConfig` also has been combined into `writeServerlessFile` - is it not called elsewhere? Also, is `writeServerlessFile` called else where which could impact it?

SiniWilson: No. As discussed it is good to go

PR #1257, opened 2020-01-07
Dedicated Roles per environment irrespective of T-Vault is enabled or not

SiniWilson:
### Requirements

Dedicated Roles per environment irrespective of T-Vault is enabled or not
### Description of the Change

Dedicated Roles per environment irrespective of T-Vault is enabled or not


builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -347,6 +348,9 @@ node() {
               lambdaARN = getLambdaARN(stackName, credsId);
               events.sendCompletedEvent('CREATE_ASSET', null, utilModule.generateAssetMap("aws", lambdaARN.arn, "lambda", config), environment_logical_id);
 
+              def iamRoleArn = utilModule.getRoleDetails(lambdaARN.arn, config['region'], credsId)
+              events.sendCompletedEvent('CREATE_ASSET', null, utilModule.generateAssetMap("aws", iamRoleArn, "iam_role", config), environment_logical_id);
```
devsatishm: It looks like that we are not creating the role as part of Jazz deployment workflow, why would we need to persist it as an asset?

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -807,11 +811,11 @@ def generateSwaggerEnv(env, deploymentNode, apiHostName, config, accountDetails)
 
 
   // TODO: the below couple of statements will be replaced with regular expression in very near future;
-  def roleId = iamRoleArnData.substring(iamRoleArnData.indexOf("::") + 2, iamRoleArnData.lastIndexOf(":"))
+  def accountId = iamRoleArnData.substring(iamRoleArnData.indexOf("::") + 2, iamRoleArnData.lastIndexOf(":"))
```
devsatishm: Isn't `accountId` available `config` already?

builds/jenkins-build-pack-api/Jenkinsfile
devsatishm: Correction - We do create it as part of serverlessDeploy before this step but shouldn't we send this event after reading the roleArn from deployment output instead of this additional call to aws lambda service?

devsatishm: Also, we make call to CloudFormation (as part of `getLambdaARN` function) previously, can we leverage that and pull other resource information during the same call?

builds/serverless-config-pack/serverless-go.yml
```diff
@@ -102,24 +102,26 @@ functions:
 #Start:resources
 resourcesDisabled:
   Resources:
+    #Start:customRoles
     #Start:isVaultEnabled
```
devsatishm: Now that we want to keep this agnostic of Vault feature being enabled or not, I would remove the word here. Instead I prefer using `isBasicRole`, `isVpcEnabled` instead of `isVaultEnabled` & `isVaultAndVpcEnabled`..

builds/jenkins-build-pack-function/Jenkinsfile
```diff
@@ -982,12 +979,20 @@ def writeServerlessFile(config, env, accountDetails) {
     sh "sed -i -- 's|{event_source_dynamodb}|${event_source_dynamodb}|g' serverless.yml"
   } 
   
-  if (config['domain'] != 'jazz' && (config['event_source_dynamodb'] || config['event_source_sqs'] || config['event_source_kinesis'] || config['event_source_s3'] 
-  || ( configLoader.TVAULT && configLoader.TVAULT.IS_ENABLED ))) {
+  if (config['domain'] != 'jazz' && !config['iamRoleARN'] ) {
```
devsatishm: This complete if-else block is little confusing. Can we add comments on when we are doing what?

builds/jazz-build-module/utility-loader.groovy
```diff
@@ -201,4 +201,19 @@ def getAzureAccountInfo(service_config){
 	return dataObj;
 }
 
+def getRoleDetails(lambdaARN, region, credsId) {
```
devsatishm: Look at my other comments, we might not need it..

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -85,6 +85,7 @@ node() {
 
       if (configObj.service_id) {
         config = serviceMetadataLoader.loadServiceMetadata(configObj.service_id)
+        config << configObj
```
devsatishm: Like it.

builds/jenkins-build-pack-api/Jenkinsfile
```diff
@@ -900,7 +900,7 @@ def writeServerlessFile(config, environment_logical_id, accountDetails){
       sh "sed -i -- '/#Start:isVaultAndVpcEnabled/,/#End:resources/d' ./serverless.yml"    
   } else  {
     sh "sed -i -- '/#Start:resources/,/#End:resources/d' ./serverless.yml"       
-    sh "sed -i -e 's|\${file(deployment-env.yml):iamRoleARN}|${iamRoleArnValue}|g' serverless.yml"
+    sh "sed -i -e 's|\${file(deployment-env.yml):iamRoleARN}|${config['iamRoleARN']}|g' serverless.yml"
```
devsatishm: Do all services with `config['domain'] == 'jazz'` have  `iamRoleARN` in their `config`? If not, this might break since the `else` block gets executed when `config['iamRoleARN']` exists OR `config['domain'] != 'jazz'`. Is that expected?

builds/delete-serverless-service-build-pack/Jenkinsfile
```diff
@@ -90,13 +90,15 @@ node {
 
 			if (configObj.service_id) {
 				service_config = serviceMetadataLoader.loadServiceMetadata(configObj.service_id)
+				service_config << configObj
```
raghits: What is the purpose of this?

SiniWilson: As discussed to make a consistency with other workflows and expecting if any future use.

builds/jenkins-build-pack-api/Jenkinsfile
SiniWilson: Done

builds/jenkins-build-pack-api/Jenkinsfile
SiniWilson: All the services with domain == 'jazz' (Core services) will have iamRoleARN in config. And it is expected.

builds/jenkins-build-pack-api/Jenkinsfile
SiniWilson: We will not get all the stack information from the "cloudformation describe-stacks" . It does not contain the role arn. So we might need the other function to get the role details.

builds/jenkins-build-pack-function/Jenkinsfile
SiniWilson: Done

builds/jazz-build-module/utility-loader.groovy
SiniWilson: Replied

builds/serverless-config-pack/serverless-go.yml
SiniWilson: Renamed

builds/jenkins-build-pack-api/Jenkinsfile
devsatishm: That's weird.. Aren't we creating these roles as part of the serverless.yml deployment (if os, it will be a part of the  same CloudFormation template)..

SiniWilson: Please see the logs in the console http://52.90.194.98:8081/job/build-pack-api/40/console

devsatishm: @SiniWilson  Should we use  [describe-stack-resources](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/describe-stack-resources.html)  instead of `describe-stacks`?

SiniWilson:
Updated the PR

PR #1258, opened 2020-01-14
Removing support for nodejs8

SiniWilson:
### Requirements

Removing support for nodejs8 runtime since it reached end-of-life date based on AWS [docs](https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html).

### Description of the Change

Removing support for nodejs8 during service creation and builds for old services using nodejs8 as runtime forcing users to use the new version.


pgobin:
**Note:** A preexisting issue with the width of the create service panel that continues to be present will be addressed in a subsequent PR after a defect is opened to track it.

PR #1259, opened 2020-01-16
Fix resource tags as per compliance requirements

devsatishm:
### Description of the Change

Using PascalCase instead of camelCase for tag keys (as per tagging requirements).

### Benefits

Resources will be provisioned as per tagging policy requirements.

### DIFF
--- a/builds/jenkins-build-pack-website/Jenkinsfile
+++ b/builds/jenkins-build-pack-website/Jenkinsfile
@@ -1064,7 +1064,9 @@ def createWebsite(stackName, resourceGroup, storageName, framework){
       storageName       : storageName,
       appName           : stackName,
       tags: [
-        "Application"   : configLoader.INSTANCE_PREFIX,
+        "Application"   : "Jazz",
+        "Platform"      : "Jazz",
+        "StackId"       : configLoader.INSTANCE_PREFIX,
         "Owner"         : service_config['created_by'],
         "Domain"        : service_config['domain'],
         "Environment"   : environment_logical_id,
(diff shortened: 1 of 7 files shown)


builds/website-distribution-pack/distribution_config.json
```diff
@@ -42,7 +42,7 @@
       "ErrorCachingMinTTL": 300
     }]
   },
-  "SmoothStreaming": true,
+  "SmoothStreaming": false,
```
raghits: Why are we changing this?

builds/jenkins-build-pack-website/Jenkinsfile
```diff
@@ -1064,12 +1064,11 @@ def createWebsite(stackName, resourceGroup, storageName, framework){
       storageName       : storageName,
       appName           : stackName,
       tags: [
-        "application"   : configLoader.INSTANCE_PREFIX,
-        "owner"         : service_config['created_by'],
-        "domain"        : service_config['domain'],
-        "STAGE"         : environment_logical_id,
-        "environment"   : environment_logical_id,
-        "service"       : stackName
+        "Application"   : configLoader.INSTANCE_PREFIX,
```
SiniWilson: In the serverless yml templates and in distribution_config_with_tags.json  "Application" mentioned as "Jazz" and here it is instance prefix. Also not every key is present here. Is that expected?

builds/website-distribution-pack/distribution_config.json
devsatishm: Unused configuration, disabling it since it won't be used in typical use cases.

builds/jenkins-build-pack-website/Jenkinsfile
devsatishm: Fixed.

PR #1266, opened 2020-02-27
support iamManagedPolicies

SiniWilson:
### Requirements

support iamManagedPolicies

### DIFF
--- a/builds/jazz-build-module/sls-app/sbr.groovy
+++ b/builds/jazz-build-module/sls-app/sbr.groovy
@@ -715,6 +715,7 @@ class SBR_Whitelist_Constraint implements SBR_Constraint {
       case "events": return whitelistValidator.validateWhitelistEvents(val); break;
       case "plugins": return whitelistValidator.validateWhitelistPlugins(val); break;
       case "actions": return whitelistValidator.validateWhitelistActions(val); break;
+      case "iamManagedPolicies": return whitelistValidator.iamManagedPolicies(val); break;
       default: throw new IllegalStateException("SBR_Whitelist_Constraint contains an unknown $elementPointer inside as follows: $val")
     }
   }
--- a/builds/jazz-build-module/sls-app/sbr.groovy
+++ b/builds/jazz-build-module/sls-app/sbr.groovy
@@ -715,7 +715,7 @@ class SBR_Whitelist_Constraint implements SBR_Constraint {
       case "events": return whitelistValidator.validateWhitelistEvents(val); break;
       case "plugins": return whitelistValidator.validateWhitelistPlugins(val); break;
       case "actions": return whitelistValidator.validateWhitelistActions(val); break;
-      case "iamManagedPolicies": return whitelistValidator.whitelistIamManagedPolicies(val); break;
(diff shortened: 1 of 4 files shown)


builds/jazz-build-module/sls-app/serverless-build-rules.yml
```diff
@@ -353,9 +353,10 @@ provider:
 
   iamManagedPolicies: # Optional IAM Managed Policies, which allows to include the policies into IAM Role
     sbr-type: "[arn-iam-policy]"
-    sbr-render: merge
+    sbr-render: user-only
```
raghits: In future if we decide to add some managed policy in config, does `merge` work as expected i.e. combine the config list with the user list? Should test with multiple config policies and multiple policies in user yml.

raghits: Here are a couple of policy ARN to whitelist so you can test it since u need at least a few more: arn:aws:iam::484695107796:policy/iam_lambda
arn:aws:iam::484695107796:policy/Lambda_Execution

SiniWilson: Added sbr-render as merge. Also tested with more policies. Working as expected.

SiniWilson:
Tested different flows and given the build rules in the story description

builds/jazz-build-module/sls-app/sbr.groovy
```diff
@@ -715,6 +715,7 @@ class SBR_Whitelist_Constraint implements SBR_Constraint {
       case "events": return whitelistValidator.validateWhitelistEvents(val); break;
       case "plugins": return whitelistValidator.validateWhitelistPlugins(val); break;
       case "actions": return whitelistValidator.validateWhitelistActions(val); break;
+      case "iamManagedPolicies": return whitelistValidator.validateWhitelistIamManagedPolicies(val); break;
```
raghits: We need to change it as a separate effort to surface the issue during validation instead of just returning true or false.

builds/jazz-build-module/sls-app/serverless-build-rules.yml
```diff
@@ -354,8 +354,11 @@ provider:
   iamManagedPolicies: # Optional IAM Managed Policies, which allows to include the policies into IAM Role
     sbr-type: "[arn-iam-policy]"
     sbr-render: merge
+    sbr-constraint:
+        sbr-whitelist: iamManagedPolicies
```
raghits: Have you validated the following cases:
1. No user managedPolicies and some config managedPolicies
2. No config managedPolicies and some user managedPolicies
3. No user managedPolicies and no config managdPolicies
4. Some user managedPolicies and some config managedPolicies (with some common and some not common)

SiniWilson: I have validated all the four cases and it works fine as expected. 

-  For 1 and 2, since no user managedpolicies, it wont come in effect because  sbr-mandatory is not set to true
- For 4 , If any outstanding it will throw error otherwise build will works
- For 3,  It will throw error as expected, since it is merge it should expect both config and userval must be there

PR #1268, opened 2020-03-27
🐛 Fix logic that identifies gitlab events correctly

devsatishm:
### Requirements

Fix logic that identifies gitlab events.

### Description of the Change

`parseInt (event.data, 10) === 0` is true whenever commit hash starts with 0 (ex: 0ade6c45bac499a6642a4fae90edeb743595fe4d). This may incorrectly identify a gitlab event as branch creation/deletion event instead of a simple commit event. 

Based on gitlab's source [code](https://gitlab.com/gitlab-org/gitlab-foss/-/blob/master/doc/api/merge_requests.md#create-mr-pipeline), it looks like `0000000000000000000000000000000000000000`  is used to identify initial/final state.

### Benefits

Correctly identifies the gitlab event type.

### DIFF
--- a/core/jazz_scm-webhook/index.js
+++ b/core/jazz_scm-webhook/index.js
@@ -241,7 +241,7 @@ var gitlabScmContextDetails = function (eventKey, body, config) {
 			var branch = origins[2];
 			result.branch = branch;
 
-			if (body.before && parseInt(body.before, 10) === 0) {
+			if (body.before && body.before === '0000000000000000000000000000000000000000') {
 				if (eventKey === 'tag_push') {
 					result.event_name = 'CREATE_TAG';
 					resolve(result);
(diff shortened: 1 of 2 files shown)


core/jazz_scm-webhook/index.js
```diff
@@ -241,7 +241,7 @@ var gitlabScmContextDetails = function (eventKey, body, config) {
 			var branch = origins[2];
 			result.branch = branch;
 
-			if (body.before && parseInt(body.before, 10) === 0) {
+			if (body.before && body.before === '0000000000000000000000000000000000000000') {
```
SiniWilson: Can we use regex for this instead checking the value?

raghits: `body.before && body.before.indexOf('0000000000') >= 0` This atleast checks there are atleast 10 zeroes in the string!

devsatishm: @raghits Highly unlikely, but what if the sha starts with 10 0s? :)

devsatishm: Closing this..

